PR #2, opened 2014-03-30
Enable the addition of extra logstash-filters through extra bosh release

mrdavidlaing:
This is an BOSH implementation showing how to integrate [additional logstash-filters](http://labs.cityindex.com/labs-team/2014/03/27/rfc-logstash-filter-repositories/) into an existing `logstash-boshrelease`

In summary, you extend your logsearch deployment manifest like this:

``` yaml
releases:
  - name: logsearch
    version: latest
  - name: logstash-filters-cf
    version: latest

...

jobs:
- name: log_parser
  release: logsearch
  templates: 
  - name: log_parser
  - name: logstash-filters-cf
    release: logstash-filters-cf
```

The `logstash-filters-cf` job places additional logstash filters in `/var/vcap/packages/logstash-filters-cf/logstash-filters-cf.conf`

On installation, the `log_parser` job includes all additional filters from `/var/vcap/packages/logstash-filters-*/logstash-filters-*.conf`

### DIFF
--- a/.final_builds/jobs/api/index.yml
+++ b/.final_builds/jobs/api/index.yml
@@ -7,4 +7,5 @@ builds:
   cfafecb25cac694ef36460c7b8785abe7ff06f0a:
     blobstore_id: f21dfcc9-53ae-4827-ac8f-ba668193a24a
     version: 2
-    sha1: f3409cf6229787d04924a6c06321ddf10cde304e
+    sha1: !binary |-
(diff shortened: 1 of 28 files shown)


dpb587:
I think this PR should have been based against the `collectd` branch instead of `master`, then we would be able to have a more precise conversation about the changes as it relates to the PR's topic of "logstash-filters". As it is now, all the extra, unrelated commits are distracting.

PR #5, opened 2014-04-10
Add logstash-forwarder

dpb587:
This adds support for forwarding the elasticsearch request logs into another deployment (like we have setup on the old, active cluster).

I suppose we could make it forward every single monit stdout/stderr log and every other vcap-related log file, but I wasn't sure if we want to do that?

http://kibana-meta-logsearch-io.monitor-cloud.cityindextest5.co.uk/index.html#dashboard/temp/QDoiH4UWR9q8VhAB8z4L7w

### DIFF
--- a/.final_builds/jobs/api/index.yml
+++ b/.final_builds/jobs/api/index.yml
@@ -7,5 +7,8 @@ builds:
   cfafecb25cac694ef36460c7b8785abe7ff06f0a:
     blobstore_id: f21dfcc9-53ae-4827-ac8f-ba668193a24a
     version: 2
-    sha1: !binary |-
-      ZjM0MDljZjYyMjk3ODdkMDQ5MjRhNmMwNjMyMWRkZjEwY2RlMzA0ZQ==
+    sha1: f3409cf6229787d04924a6c06321ddf10cde304e
+  858302bf14826a83697bb70a27200d5092263806:
+    blobstore_id: bad8d0b5-c479-4763-84b9-f312b99bfd68
+    version: 3
+    sha1: c32c2af121f05f7eff03bfa252c16db3568d4a76
--- a/examples/bosh-lite.yml
+++ b/examples/bosh-lite.yml
@@ -49,6 +49,14 @@ jobs:
     elasticsearch:
       node:
         allow_data: false
+    logstash_forwarder:
+      files:
+      - path: /var/vcap/sys/log/elasticsearch/requests.log
+        type: elasticsearch_request
+      - path: /var/vcap/sys/log/api/nginx.access.log
+        type: nginx_access
+      - path: /var/vcap/sys/log/api/nginx.error.log
+        type: nginx_error
 
 - name: ingestor
   release: logsearch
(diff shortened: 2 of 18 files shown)


dpb587:
@mrdavidlaing, thoughts and review comments?

dpb587:
@mrdavidlaing, I committed changes for the proposal, but had two problems.
1. You were right that the file-based key name didn't work with bosh, so it's what you originally said.
2. I tried putting both global and job-specific log paths, but the job-specific ones were the only ones which got applied. I'm not sure recursive merging works in this particular way. Feel free to take a look and tell me it actually does work and how :)

I included the logstash paths in the `bosh-lite.yaml`, but it does not include the `logstash_forwarder` job (since it requires host/cert config).

mrdavidlaing:
@dpb587 - I've asked the bosh-users mailing list for clarrification on global + job property merging

https://groups.google.com/a/cloudfoundry.org/d/msg/bosh-users/AGkpihXWzOY/MRzEkIKGCQEJ

dpb587:
The mailing list confirms the global/job property merging I saw as expected behavior.

mrdavidlaing:
Bummer.

I guess if we leave it as is you can choose to use global OR job properties, which has some value.

When overriding you would be forced to respecify all paths.

dpb587:
Alternatively, we have two properties: `logstash_forwarder.global_paths` and `logstash_forwarder.job_paths`. Confusing or worthwhile?

mrdavidlaing:
Worthwhile.  Lets do that

aristotelesneto:
Is logstash_forwarder lumberjack? If so I'd have switched the naming to: for the job simply "forwarder"; for the package the job uses simply "lumberjack" in order to maintain consistency with the other names.

Although I'd switch log_parser to just parser, just to take care of my itching OCD :D

dpb587:
The forwarder [project](https://github.com/elasticsearch/logstash-forwarder) was initially and simply named `lumberjack`, but semi-recently transitioned to being called `logstash-forwarder`. The protocol maintains the name `lumberjack`, but the program we're using to forward logs is now called `logstash-forwarder`. From that perspective, I think the naming used here is consistent with the external concepts that people may already be familiar with, but we can reconsider if there's a better approach for us and logsearch users.

You might be right on the `log_parser` vs `parser` thing... hmm. @mrdavidlaing, thoughts on whether we can/should rename `log_parser` to `parser`?

aristotelesneto:
I see, didn't realise lumberjack had been renamed. In that case, you're absolutely right, the package should be named 'logstash-forwarder'.

I'd still maintain that the logsearch bosh release job, however, should be named 'forwarder' as that would more accurately reflect bosh jobs (which can be seen as roles, which are collection of one or more packages, irrespective of each package name) and ties in better with the rest of the release - we'd have forwarder, ingestor, queue, parser, search/store, api.

Just my 2c.

mrdavidlaing:
>  thoughts on whether we can/should rename log_parser to parser?

Sorry for being so late to the party.  A PR renaming `log_parser` to `parser` is going to get a :+1: from me

PR #7, opened 2014-04-15
Switch to TCP/UDP inputs instead of syslog

aristotelesneto:
The syslog input causes _grokparsfailures due to CF messages not matching the
standard linux-syslog message format. Because of the failure, CF rules do not
come into effect.

https://github.com/elasticsearch/logstash/blob/1.3.x/lib/logstash/inputs/syslog.rb#L55-L69
https://github.com/logsearch/logstash-filters-cf/blob/master/src/100-cloudfoundry.conf.erb#L8


mrdavidlaing:
@wdneto - Thanks for this patch!

I'm a little confused as to why you are seeing this error and not me.

Could you share your CF `syslog_aggregator` config so I can try replicate the issue.

mrdavidlaing:
I'm using this:

```
  syslog_aggregator:
    address: ingestor.meta.logsearch.io 
    port: 5515
    transport: relp
```

which is actually going to the relp_ingestor rather than the syslog_ingestor

aristotelesneto:
I'm running v166 of cf-release - so there might have been changes between the version I'm running and yours. Based on your video, one definite difference is that I'm using transport: tcp and you're using transport: relp.

Heres the forwarder from the cc-api job: https://gist.github.com/wdneto/5065f50c69a84f0367ec

Based on the logstash syslog.rb file, though, the issue seems to be entirely related to a missing ":".

https://github.com/elasticsearch/logstash/blob/1.3.x/lib/logstash/inputs/syslog.rb#L58
"SYSLOGLINE", :pattern=>"%{SYSLOGBASE2} %{GREEDYDATA:message}"
"SYSLOGBASE2", :pattern=>"(?:%{SYSLOGTIMESTAMP:timestamp}|%{TIMESTAMP_ISO8601:timestamp8601}) (?:%{SYSLOGFACILITY} )?%{SYSLOGHOST:logsource} %{SYSLOGPROG}:"

Which I think is failing because of the ":" at the very end of the pattern, resulting in a _grokparsefailure tag even before the message gets to redis.

Additionally to switching to plain tcp/udp inputs, I'd also recommend tagging grok with a different tag to the default, using 'tag_on_failure', then using that tag to check (happy to send a pr with this mod aswell).

mrdavidlaing:
> Additionally to switching to plain tcp/udp inputs, I'd also recommend tagging grok with a different tag to the default, using 'tag_on_failure', then using that tag to check (happy to send a pr with this mod aswell).

Good idea.  Please send PR

jobs/ingestor_syslog/templates/config/syslog_to_redis.conf.erb
```diff
@@ -1,5 +1,10 @@
 input {
-	syslog {
+	tcp {
```
mrdavidlaing, line 2: I think [this comment in logstash is revealing](https://github.com/elasticsearch/logstash/blob/1.3.x/lib/logstash/inputs/syslog.rb#L17) 

```
Of course, 'syslog' is a very muddy term. This input only supports RFC3164
```

With the change you are proposing; we're actually just accepting anything on a TCP/UDP socket, and then calling it `syslog`

I think we should actually be adding a new `ingestor_tcp`, and possibly rename this one to `ingestor_syslog_RFC3164`

What do you think?

mrdavidlaing, line 2: `ingestor_tcp` would tag logs as `type = tcp`

aristotelesneto, line 2: That would work. As an alternative solution, we could do the same as what the syslog input does, at the logstash-filters-common level.

This would allow you to still support both methods (but allowing you to put if statements around the rfc3164 checks), and still allows you to specify syslog_forwarder.all on CF, thereby capturing both vcap as well as general server syslog (which means you cannot have multiple log endpoints).

As for naming, could we not simply merge all ingestors into one, and allowing ingestor.transport: (relp|tcp|udp|lumberjack)?

aristotelesneto:
Tag check sent: https://github.com/logsearch/logstash-filters-cf/pull/1

jobs/ingestor_syslog/templates/config/syslog_to_redis.conf.erb
mrdavidlaing, line 2: > As an alternative solution, we could do the same as what the syslog input does, at the logstash-filters-common level.

That is a better idea.  PR? :)

mrdavidlaing, line 2: > merge all ingestors into one

I'd prefer to keep them separate for now.

aristotelesneto, line 2: Sure, I'll add the syslog parsing later today. I'm assuming we'd maintain the name ingestor_syslog - otherwise it'd have to be split between ingestor_tcp/udp?

mrdavidlaing, line 2: Yes, lets keep `ingestor_syslog`

aristotelesneto:
Replaced by #12, so I can delete my fork.

PR #8, opened 2014-04-15
Support customizing elasticsearch environment and options

dpb587:
This would allow us to adjust elasticsearch runtime options (e.g. garbage collection flags) and environment variables (e.g. heap and paths).

### DIFF
--- a/examples/bosh-lite.yml
+++ b/examples/bosh-lite.yml
@@ -146,20 +146,6 @@ properties:
     exec:
       environment:
         ES_HEAP_SIZE: 256M
-      options:
-      - "-XX:+UseParNewGC"
-      - "-XX:+UseConcMarkSweepGC"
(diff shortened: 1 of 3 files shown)


dpb587:
@mrdavidlaing, thoughts or review comments before merging?

mrdavidlaing:
@dpb587 I think there is real value in us shipping with the sane default settings which we have discovered the hard way.

Can you modify this so that:
1.  We use our default elasticsearch runtime options if none are specified in the deploy manifest
2.  The amount of ES heap is dynamic if not specified

dpb587:
@mrdavidlaing, I've made the adjustments. I think in most cases we'll need to copy/override the JVM options with instance-specific ones, but now most of our defaults are there which should at least provide anyone with a head start to debugging and creating their own appropriate settings.

mrdavidlaing:
Perfect.  Lets roll this into a release

PR #14, opened 2014-04-24
Add kibana as plugin

mrdavidlaing:
Its helpful to have a "plain" kibana installed under `api.{your-dns}/_plugin/kibana`


dpb587:
:thumbsup:

PR #15, opened 2014-04-24
Fix debug property

dpb587:
Since `if_p` seems to be more for property presence rather than truthyness.


dpb587:
@mrdavidlaing, care to review and merge?

mrdavidlaing:
I am Mr Merge today!

PR #17, opened 2014-04-25
Fix elasticsearch mapping being ignored

mrdavidlaing:


jobs/elasticsearch/templates/config/templates/logstash_default.json.erb
```diff
@@ -28,7 +28,7 @@
                 "_ttl" : { 
                     "enabled" : true,
                     "default" : "<%= p("elasticsearch.indices.ttl_default") %>"
-                }
+                },
```
mrdavidlaing, line 31: Nice catch!

mrdavidlaing:
Note that these fixes will only get applied to new daily indexes

Issue #9, opened 2014-04-17
Better way to integrate filters


aristotelesneto:
Hi David,

Would you be interested if I went away and formulated a better way to integrate the filters into the main logsearch-boshrelease, so we don't need to have a separate filter bosh release?

I was thinking about something along the lines of leveraging the pre_packaging / packaging scripts, along with a job that (if configured to do so), would periodically check that the filters are up to date with a remote git repo (from the deployed machine).

Thoughts?

mrdavidlaing:
YES!

What we currently have is kinda crummy; so any improvements to the flow much appreciated.

The design goals are:
1.  Deploying logsearch-boshrelease gets you set the set of https://github.com/logsearch/logsearch-filters-common filters
2.  You should then be able to "extend" your logsearch deployment by choosing to add some community filters - eg, https://github.com/logsearch/logstash-filters-cf
3.  To add your own custom filters, you should fork https://github.com/logsearch/logsearch-filters-common, replace the common with your own filters and then "extend" your logsearch deployment with those as-well.

In our experience, once you have your logsearch cluster running, updating filters is by _far_ the most common change.  Ideally we want an easily "continuous deployable" way to (a) edit `logsearch-filters-mine`, (b) run unit tests, and then (c) "bosh deploy" them into your existing cluster.

mrdavidlaing:
@wdneto - I've given you commit access to the relevant repo's - please just make your PR branch in the repo itself.  Makes it easier for us to collaborate on PRs.

Please _don't_ merge to master

aristotelesneto:
Cool, thanks. Will make sure the branch is in the logsearch repo.

Upon reading the design goals, the first thing that came to mind was the way buildpacks are now installed into cf as of v166, except without the whole api side of things. 

I'll have a play and see what I can come up with. 

Did you have any thoughts on a self-updating job or would you rather new rules still be pushed via bosh deploy (rather than pulled by the already deployed machine)?

mrdavidlaing:
> Did you have any thoughts on a self-updating job or would you rather new rules still be pushed via bosh deploy (rather than pulled by the already deployed machine)?

@dpb587 and I are against a self updating job; since it makes auditing the current state of the system difficult.  I think a solution that requires a `bosh deploy` to be run is the right way to go.

mrdavidlaing:
@wdneto Do you have any code / designs to share for this?

dpb587:
[...]

PR #19, opened 2014-04-30
Job definition filter implementation

dpb587:
This is an implementation of what @mrdavidlaing proposed in his [comment](https://github.com/logsearch/logsearch-boshrelease/issues/9#issuecomment-41587591).

Users are able to update the `logstash_parser.filters` config with URLs of their tar/tar.gz/zip files. Files are only downloaded once and are extracted in order of appearance within the array value. This means a later file could overwrite an earlier, upstream version if using the exact same name (which I consider a feature). Once extracted, they're loaded into the logstash configuration file still in order of priority based on the name.


jobs/log_parser/spec
```diff
@@ -2,7 +2,6 @@
 name: log_parser
 packages: 
   - logstash
-  - logstash-filters-common
```
mrdavidlaing, line 5: @dpb587 I think that the common filters should be part of the base boshrelease; so you don't have an external dependancy "out the box"

dpb587:
So we're going to maintain both boshrelease filter configuration _and_ job filter configuration?

mrdavidlaing:
No.  Let me show you what I mean

mrdavidlaing:
Relocated to target `develop`

Issue #18, opened 2014-04-25
Formalise process for creating new releases


mrdavidlaing:
What should the branching process for creating a final release be?

We want to optimise for continuous delivery - ie, we want regular small changes to be deployed rather than batching them up into periodic releases.

We have third party consumers; who might not want quite so many releases; and definately want to know what changes are in each release.

mrdavidlaing:
I'm wondering whether Git Flow might not be part of the answer.
1.  `feature-branch` gets new feature
2.  `feature-branch`-PR gets discussion (and possibly an automated build / deploy / test run)
3.  merge to `develop` branch - do your continuous deployment from here, creating and deploying your own `dev-release`s
4.  Weekly commit a new release to `develop`, create a PR to merge to `master` that contains the release notes/changelog (sort of like https://github.com/logsearch/logsearch-boshrelease/pull/16)
5.  Third parties consume from `master`

mrdavidlaing:
@wdneto, @dpb587 - comments?

dpb587:
What do other bosh repositories do for their processes?

mrdavidlaing:
https://github.com/cloudfoundry/cf-release seems to be the "canonical" BOSH release

They have a `develop` branch; into which [PRs are merged](https://github.com/cloudfoundry/cf-release/pulls?direction=desc&page=1&sort=created&state=closed) 

Those PRs [have the uploaded blobs, but no release](https://github.com/cloudfoundry/cf-release/pull/366/files#diff-0)

I'm kinda confused as to what happens after that.

aristotelesneto:
After I started messing with CF, I did enjoy learning about the Pivotal processes. The cf-release have the following branches which are involved in their release pipeline:

```
The pipeline interacts with the following ‘cf-release’ branches:
develop: Developers commit to this branch;
runtime-deployed: Continuous Integration system (CI) deploys this branch to Test environment (tabasco);
runtime-passed: CI runs Cloud Foundry Acceptance Tests (CATs) in Test environment and then promotes;
acceptance-deployed: CI deploys to Staging environment (a1) from this branch;
release-candidate: CI runs CATs in Staging environment and then promotes;
deployed-to-prod: Deploy to Production Environment (pws) from this branch;
master: Promoted (tagged) what was successfully deployed to Production;
```

They then leverage tags for when releases are created from the blobs / ready for world consumption, and promoted to master branch and individual commit progress tracked via http://git-pipeline.cfapps.io/

[...]

PR #20, opened 2014-05-01
job definition filter implementation /issue #9

mrdavidlaing:
This is an implementation of what @mrdavidlaing proposed in his [comment](https://github.com/logsearch/logsearch-boshrelease/issues/9#issuecomment-41587591).

Users are able to update the `logstash_parser.filters` config with URLs of their tar/tar.gz/zip files. Files are only downloaded once and are extracted in order of appearance within the array value. This means a later file could overwrite an earlier, upstream version if using the exact same name (which I consider a feature). Once extracted, they're loaded into the logstash configuration file still in order of priority based on the name.

---

Relocated from https://github.com/logsearch/logsearch-boshrelease/pull/19 to target `develop` as per #18


packages/logstash/spec
```diff
@@ -3,3 +3,4 @@ name: logstash
 dependencies: []
 files: 
   - logstash/logstash-1.3.0.jar
+  - logstash/logsearch-filters-common--21e1a3b643.tgz
```
mrdavidlaing, line 6: Package logsearch-filters-common with the bosh release

examples/bosh-lite.yml
```diff
@@ -158,6 +158,8 @@ properties:
     host: 10.244.2.10
   logstash_parser:
     debug: true
+    filters:
+    - "https://ci-logsearch.s3.amazonaws.com/logsearch-filters-cf/logstash-filters-cf-7.tgz"
```
mrdavidlaing, line 162: Also allow pulling extra external filters

jobs/log_parser/templates/bin/log_parser_ctl
```diff
@@ -20,17 +20,60 @@ case $1 in
...
+    pushd ${JOB_DIR}/config > /dev/null
+
+    <% p('logstash_parser.filters').each_with_index do | v, i | %>
+      # <%= v %>
+      if ! [[ -d raw/<%= md5filters[i] %> ]] ; then
+        mkdir -p raw/<%= md5filters[i] %>
+        pushd raw/<%= md5filters[i] %> > /dev/null
+
+        <% if v =~ /\.tar\.gz$/ or v =~ /\.tgz$/ %>
+          /usr/bin/wget -qO- "<%= v %>" | /bin/tar -xz
```
mrdavidlaing, line 52: I dropped the `--strip-components 1` to support filters that don't come packaged inside a subfolder

jobs/log_parser/templates/bin/log_parser_ctl
```diff
@@ -20,17 +20,60 @@ case $1 in
...
+        <% elsif v =~ /\.tar$/ %>
+          /usr/bin/wget -qO- "<%= v %>" | /bin/tar -x 
+        <% elsif v =~ /\.zip$/ %>
+          /usr/bin/wget "<%= v %>" && /usr/bin/unzip -d .ziptmp `basename "<%= v %>"` && /bin/mv .ziptmp/*/* ./ && /bin/rm -fr `basename "<%= v %>"` .ziptmp
+        <% end %>
+
+        popd > /dev/null
+      fi
+
+      cp -R raw/<%= md5filters[i] %>/* composite/
```
mrdavidlaing: The `-R` should cover filters that are nested within subfolders

dpb587:
Okay. I think there may be problems if multiple filtersets (package or job) have a file by the same name (they may be applied inconsistently due to file system listings and sort only using the base file name).

Issue #51, opened 2014-07-01
Zero downtime ES deployments


mrdavidlaing:
@dpb587 asks

> How can we adjust our bosh deployments to upgrade/restart elasticsearch nodes one at a time (serial = true), but then wait for green cluster before continuing with the next node?

mrdavidlaing:
@wdneto - do you have any ideas here?

aristotelesneto:
This can be accomplished by combining a couple of things together.

1) Override the global "update:" on the job you only want 1 node to be restarted at a time on the deployment manifest, and set "max_in_flight: 1". This ensures only one node is ever worked on at a time.

2) Setup a drain script for the job.  The way drain scripts work, is that before bosh sends a shutdown command it runs the drain script, meaning that it won't prevent BOSH from continuing to the next node per se, but instead it will prevent the next node from going down before knowing the cluster is ok (ie. the script runs on the next node, not on the node you just upgraded).

Simply doing option 2 should suffice, but I think it's probably best to do both.

The drain script isn't really documented, but [here is where BOSH handles it](https://github.com/cloudfoundry/bosh/blob/master/bosh_agent/lib/bosh_agent/message/drain.rb). 
A couple of major quirks:
- The sample script has to be named "bin/drain" (destination file on the job spec)
- If you're collocating multiple jobs together, the drain script will only run for the first job specified in the manifest (bug, reported but not yet fixed) 
- The drain script expects only an integer as output (on stdout, any other text will cause it to error), as well as an exit code of 0. 
  - Exiting with anything other than 0 means you will need to ssh into the machine to fix the reason why (ie. theres no command you can send to bosh from the outside to tell it to ignore the exit code of the drain script).
  - If the output of the script is a negative integer (eg: -N) , bosh wait N seconds before sending another drain command, however, when it does the arguments of the script changes, which I found very annoying.

Here is [a basic sample](https://gist.github.com/wdneto/b6f64c805652ddfb60ca) I used for removing an OSD from a Ceph cluster when a node is deleted.

The useful available options are job_shutdown (when the node is being deleted, useful for removing itself from a cluster, triggering permanent evacuation) and job_changed (job is being updated, but should still remain in the cluster).

aristotelesneto:
[Here is another sample drain script](https://github.com/cloudfoundry/cf-release/blob/master/jobs/dea_next/templates/deterministic_drain.rb), from the DEA.

mrdavidlaing:
[...]

PR #56, opened 2014-07-07
Add drain script for elasticsearch

dpb587:
This is the continuation and implementation of issue #51 with concrete changes to discuss. The goal is to support zero downtime of the elasticsearch cluster during deploys (which has been a significant problem).


dpb587:
I consider this to be a sufficiently safe and functional iteration to merge into master (drain script is still disabled by default in the spec).

dpb587:
Merged via 151fe2b51f050a3b318c1300472327baa7b6b831

Issue #58, opened 2014-07-07
Reconsider how we do logstash filters


dpb587:
I think our current approach to logstash filter configuration is too fragile. I think `logsearch-boshrelease` should manage the architecture and define the conventions, but have minimal impact on its usage. While the set of filters in `logsearch-filters-common` is convenient to have pre-installed, it becomes very inconvenient to end deployments because it significantly affects how users' events are parsed and affects dashboards that we don't already manage and provide an upgrade path for.

For example, we updated `json` in common, but we also had a `json` internally with our own timezone configuration and another patch. If we deployed release 15 because we wanted the upgraded dependencies, we would be breaking how all our events are parsed because the common `json` filter was configured with a higher precedence than our internal one. Even knowing that difference, we're still hard pressed for getting an integration test suite going and being confident in the upgrade.

I propose...
1. We no longer embed `logsearch-filters-common` in our `log_parser`. We can default a published `logsearch-filters-common` artifact as the default `logstash_parser.filters` in the `spec`, but we don't forcefully embed the filters.
2. We refactor the logstash config-generating script out of `log_parser_ctl` and into a separate script which accepts arbitrary arguments of filters to use (e.g. `*.tar`, `*.tar.gz`, `*.zip`, `file://*`) and outputs to `STDOUT` the usable logstash `filters { ... }` section.
3. We add an "integration test" type command to `logsearch-filters-common` which reuses that refactored script, allowing end users to write integration tests against a known set of filters.
4. We require users to include the `logsearch-filters-common` artifact alongside their own artifacts in their deployment manifest, if they want to continue using it.

Which means, most notably, that users can upgrade the architecture and the already user-customizable parser side of things independently and safely.

dpb587:
@mrdavidlaing, @wdneto thoughts?

mrdavidlaing:
`logsearch-filters-common` currently contains the following:

**Transport**:
1. 10-syslog_standard.conf
2. 20-nxlog_standard.conf
**Data**:
3. 30-json.conf
4. 75-elasticsearch_request.conf
5. 75-nginx_combined.conf

I think we should remove all but the Transport filters from  `logsearch-filters-common` and then externalised as suggested.

dpb587:
I find it difficult to draw the line between transport and message format. For example, nxlog could use the syslog protocol to ship syslog messages (i.e. double encoded), but after nxlog, there's not a syslog parser.

[...]

PR #60, opened 2014-07-09
Adjust logstash.filters configuration

dpb587:
This includes a few changes initiated by logsearch/logsearch-boshrelease#58...
- by default, embedded filters are no longer included if you specify your own filters (`example/warden.yml` provides an example to reuse them within a deployment)
- this adds support for `file://*` which will include `*.conf` files from the specified directory
- this simplifies the config building a little bit and allows `logstash.filters` to be either a string or array of location paths (if an array, files of the same name will overwrite earlier extractions)


dpb587:
Merged via e91ea4af7b7b664c85725b1b235891744f808be4

PR #61, opened 2014-07-10
Add CI tasks that can be run in a Docker container

mrdavidlaing:
Enables a Docker based build server to create logsearch dev releases using a command like:

```
docker build -t $JOB_NAME $(pwd)

docker run -e BOSH_TARGET=https://xx.yy.zz.bb:25555 -e BOSH_USER=sudobot -e BOSH_PASSWORD=$SUDOBOT_BOSH_PASSWORD -v $(pwd):/workspace -t $JOB_NAME rake dev_release:create_and_upload_and_deploy
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -15,7 +15,7 @@ namespace :dev_release do
 	desc "Creates DEV release"
 	task :create do
 		puts "=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-="
-		sh "bosh #{target} create release --force"
+		sh "yes logsearch | bosh create release --force --with-tarball"
 	end
 	desc "Creates and uploads DEV release to currently targeted BOSH."
 	task :create_and_upload => :create do
--- a/Rakefile
+++ b/Rakefile
@@ -5,18 +5,27 @@ RSpec::Core::RakeTask.new('spec')
 # If you want to make this the default task
 task :default => :spec
 
+target=""
+target="#{target} -t #{ENV['BOSH_TARGET']}" if ENV['BOSH_TARGET']
+target="#{target} -d #{ENV['BOSH_DEPLOYMENT']}" if ENV['BOSH_DEPLOYMENT']
+target="#{target} -u #{ENV['BOSH_USER']}" if ENV['BOSH_USER']
+target="#{target} -p #{ENV['BOSH_PASSWORD']}" if ENV['BOSH_PASSWORD']
+
 namespace :dev_release do
+	desc "Creates DEV release"
+	task :create do
+		puts "=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-="
+		sh "bosh #{target} create release --force"
+	end
 	desc "Creates and uploads DEV release to currently targeted BOSH."
-	task :create_and_upload do
+	task :create_and_upload => :create do
 		puts "=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-="
-		sh "bosh create release --force"
-		puts "=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-="
-		sh "bosh -n upload release"
+		sh "bosh #{target} -n upload release"
 	end
 	desc "Creates, uploads and deploys DEV release to currently targeted BOSH deployment manifest"
 	task :create_and_upload_and_deploy => :create_and_upload  do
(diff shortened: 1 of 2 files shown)


Rakefile
```diff
@@ -5,18 +5,27 @@ RSpec::Core::RakeTask.new('spec')
...
+target="#{target} -p #{ENV['BOSH_PASSWORD']}" if ENV['BOSH_PASSWORD']
+
 namespace :dev_release do
-	desc "Creates and uploads DEV release to currently targeted BOSH."
-	task :create_and_upload do
+	desc "Creates DEV release"
+	task :create do
 		puts "=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-="
-		sh "bosh create release --force"
+		sh "yes logsearch | bosh create release --force --with-tarball"
```
dpb587, line 18: `yes` is an infinite command, isn't it? Couldn't we just use the affirmative `-n` flag to bosh?

Rakefile
```diff
@@ -5,18 +5,27 @@ RSpec::Core::RakeTask.new('spec')
 # If you want to make this the default task
 task :default => :spec
 
+target=""
+target="#{target} -t #{ENV['BOSH_TARGET']}" if ENV['BOSH_TARGET']
+target="#{target} -d #{ENV['BOSH_DEPLOYMENT']}" if ENV['BOSH_DEPLOYMENT']
+target="#{target} -u #{ENV['BOSH_USER']}" if ENV['BOSH_USER']
+target="#{target} -p #{ENV['BOSH_PASSWORD']}" if ENV['BOSH_PASSWORD']
```
dpb587, line 12: Nice addition

Rakefile
mrdavidlaing, line 18: No; when creating a dev release for the first time (ie, on a clean buiildserver repo) we have to specify a release name

dpb587, line 18: I just tested it locally by trashing my `dev_releases` and `config/dev.yml` and running `bosh create release --force --with-tarball`. Without `-n` it asks, but with `-n` it defaults to `logsearch` and creates the release anyway. I think this line could switch to `-n`?

dpb587, line 18: Another question - by default won't the build server always be re-creating a dev release as version 1 since it's not saving state anywhere? Would it be better to use `ENV['BUILD_NUMBER']` in the dev release version, when available?

aristotelesneto, line 18: Why not just pre-populate config/dev.yml and run with -n?

Alternatively to using 'yes', one could also use 'printf' (which isn't infinite), and has worked for me on occasions when yes wouldn't.

dpb587, line 18: I like the pre-populated `config/dev.yml` idea.

mrdavidlaing, line 18: If would be awesome if we could use a pre-populated config/dev.yml and run with -n to get

`logsearch.{ENV['BUILD_NUMBER']}-dev` (or something similar that links the build to the dev release)

Could one of you submit a commit that makes that work?

dpb587, line 18: Unfortunately, it looks like we can't manually specify versions when using `--force`...

```
$ bosh create release --force --with-tarball --version build147
...snip...
Version numbers cannot be specified for dev releases
```

I suppose we could let it build whatever version it wanted, then manually untar, patch release.MF, retar the build artifact; but I'm not sure it's worth it?

Also, what are going to be happening with these dev builds? Also does this mean I must immediately upload blobs whenever I commit a change, even if I'm not entirely sure it will be in the final merge? Otherwise it would cause build failures, right?

mrdavidlaing, line 18: > Also, what are going to be happening with these dev builds?

I'd like the build servers (dev) release TGZ to be published to S3, so if we're happy with it we / others can

```
bosh upload release https://s3...../logsearch-boshrelease-16.3-dev.tgz
bosh deploy
```

on their infrastructure(s), getting the same release numbers

Concerning uploaded blobs - yes; at the point you are happy with the blobs on your bosh-lite, and want to commit to `develop` to have the CI system create & deploy a (dev) release to our testing AWS cluster.

Its a similar workflow to "run the unit tests on my machine", then commit and have the build server "run the tests on some standardised hardware".  It we end up with some additional blobs in S3; then that is a worthwhile price to pay.

dpb587:
@mrdavidlaing, this is a (very) long running PR with no recent activity. I'm not using the existing Dockerfile or Rakefile myself - is this something we currently need and should still be merged, or shall we close it?

mrdavidlaing:
We should close it without merging in favour of investigating a consol.io
approach

On Wednesday, May 6, 2015, Danny Berger notifications@github.com wrote:

> @mrdavidlaing https://github.com/mrdavidlaing, this is a (very) long
> running PR with no recent activity. I'm not using the existing Dockerfile
> or Rakefile myself - is this something we currently need and should still
> be merged, or shall we close it?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/logsearch/logsearch-boshrelease/pull/61#issuecomment-99566368
> .

## 

David Laing
logsearch.io - build your own open source cloud logging cluster
http://davidlaing.com

Issue #63, opened 2014-07-15
syslog_tls should not require configuration


dpb587:
Currently it errors with the following if configuration properties are not provided...

```
Error 100: Error filling in template `syslog_tls.crt.erb' for `ingestor/0' (line 1: Can't find property `["logstash_ingestor.syslog_tls.ssl_cert"]')
Error 100: Error filling in template `syslog_to_redis.conf.erb' for `ingestor/0' (line 15: Can't find property `["logstash_ingestor.syslog_tls.port"]')
```

Instead, I believe the `syslog_tls` service should simply not run when the properties are empty.

mrdavidlaing:
@dpb587,

>  the syslog_tls service should simply not run when the properties are empty.
> This is the correct behaviour.  Sorry about the bug.

PR #64, opened 2014-07-15
Make syslog_tls support optional, enabled by the port property

dpb587:
Issue #63


dpb587:
@mrdavidlaing, are you okay with this change?

mrdavidlaing:
>  the syslog_tls service should simply not run when the properties are empty.

This is the correct behaviour.  

Please merge

PR #80, opened 2014-10-27
Add archive restoration capabilities

dpb587:
This PR adds the scripts for restoring data from one or more archives. The `examples/bosh-lite.yml` includes an example job, `restorer`, which is pre-configured to both ingest and parse restored messages. If you need rapid restoration, you'll probably want to start additional instances. Note that in most cases, you will want to use a separate restoration queue in redis (otherwise you'll end up double-archiving these restored events).

The following commands are also included in the `examples/bosh-lite.yml` above the `restorer` job:

```
# pipe our local file to a remote tmp file, sudo, pipe that tmp file into the command, delete our tmp file; kind of convoluted
$ cat 2014102715.20141027150019-api-0.log.xz | ssh -i "$HOME/.ssh/vagrant" vcap@10.244.2.26 "cat > /tmp/ssh-$$ && echo c1oudc0w | sudo -p '' -S -- /bin/bash -c 'cat /tmp/ssh-$$ | /var/vcap/jobs/ingestor_archiver/bin/reload xz' && rm /tmp/ssh-$$"
Stopping logstash...done
Purging existing logs, state...done
Transferring logs...done
Extracting logs....done
Starting logstash...............done

# run the utility script which blocks until we've read all the waiting log files
$ ssh -i "$HOME/.ssh/vagrant" vcap@10.244.2.26 "echo c1oudc0w | sudo -p '' -S -- /var/vcap/jobs/ingestor_archiver/bin/wait-until-loaded"
/var/vcap/data/archiver-restorer/restore/_reload: 1623006 of 923651191
/var/vcap/data/archiver-restorer/restore/_reload: 16394000 of 923651191
/var/vcap/data/archiver-restorer/restore/_reload: 36017618 of 923651191
/var/vcap/data/archiver-restorer/restore/_reload: 47640964 of 923651191
/var/vcap/data/archiver-restorer/restore/_reload: 61461106 of 923651191
/var/vcap/data/archiver-restorer/restore/_reload: 78232168 of 923651191
/var/vcap/data/archiver-restorer/restore/_reload: 102429712 of 923651191
...snip...
/var/vcap/data/archiver-restorer/restore/_reload: 919588642 of 923651191

$ echo $?
0
```

### DIFF
--- a/examples/bosh-lite.yml
+++ b/examples/bosh-lite.yml
@@ -148,11 +148,11 @@ jobs:
         allow_master: false
 
 # upload your backfill data
-# $ scp -i "$HOME/.ssh/vagrant" 2014102715.20141027150019-api-0.log.xz vcap@10.244.2.26:/tmp/backfill.xz
+# $ sshpass -pc1oudc0w scp -i "$HOME/.ssh/vagrant" 2014102721.20141027220003-api-0.log.xz vcap@10.244.2.26:/tmp/backfill.xz
 # restart ingestion of it all
-# $ ssh -i "$HOME/.ssh/vagrant" vcap@10.244.2.26 "echo c1oudc0w | sudo -p '' -S -- /bin/bash -c '/var/vcap/jobs/ingestor_archiver/bin/reload /tmp/backfill.xz'"
+# $ sshpass -pc1oudc0w ssh -i "$HOME/.ssh/vagrant" vcap@10.244.2.26 "echo c1oudc0w | sudo -p '' -S /var/vcap/jobs/ingestor_archiver/bin/reload /tmp/backfill.xz"
 # utility if you want to block until complete
-# $ ssh -i "$HOME/.ssh/vagrant" vcap@10.244.2.26 "echo c1oudc0w | sudo -p '' -S -- /var/vcap/jobs/ingestor_archiver/bin/wait-until-loaded"
+# $ sshpass -pc1oudc0w ssh -i "$HOME/.ssh/vagrant" vcap@10.244.2.26 "echo c1oudc0w | sudo -p '' -S /var/vcap/jobs/ingestor_archiver/bin/wait-until-loaded"
 - name: restorer
   release: logsearch
   templates:
--- a/examples/bosh-lite.yml
+++ b/examples/bosh-lite.yml
@@ -147,8 +147,12 @@ jobs:
       node:
         allow_master: false
 
-# cat 2014102715.20141027150019-api-0.log.xz | ssh -i "$HOME/.ssh/vagrant" vcap@10.244.2.26 "cat > /tmp/ssh-$$ && echo c1oudc0w | sudo -p '' -S -- /bin/bash -c 'cat /tmp/ssh-$$ | /var/vcap/jobs/ingestor_archiver/bin/reload xz' && rm /tmp/ssh-$$"
-# ssh -i "$HOME/.ssh/vagrant" vcap@10.244.2.26 "echo c1oudc0w | sudo -p '' -S -- /var/vcap/jobs/ingestor_archiver/bin/wait-until-loaded"
+# upload your backfill data
+# $ scp -i "$HOME/.ssh/vagrant" 2014102715.20141027150019-api-0.log.xz vcap@10.244.2.26:/tmp/backfill.xz
+# restart ingestion of it all
+# $ ssh -i "$HOME/.ssh/vagrant" vcap@10.244.2.26 "echo c1oudc0w | sudo -p '' -S -- /bin/bash -c '/var/vcap/jobs/ingestor_archiver/bin/reload /tmp/backfill.xz'"
+# utility if you want to block until complete
+# $ ssh -i "$HOME/.ssh/vagrant" vcap@10.244.2.26 "echo c1oudc0w | sudo -p '' -S -- /var/vcap/jobs/ingestor_archiver/bin/wait-until-loaded"
 - name: restorer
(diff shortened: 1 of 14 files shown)


dpb587:
We may want to discuss the current implementation of how files are cat'd across SSH. I generally like this approach, but the sudo requirements make it kind of convoluted. Having submitted this, now I think I should try temporarily renaming stdin/fd0 to avoid the tmp file to keep that simple approach. Alternatively, we simply change the script to assume arguments are local paths that it needs to process and that the user will have uploaded them in a previous step.

mrdavidlaing:
@dpb587, Could you walk through how this would be used in a live cluster running on AWS.

Assume the following scenario:
1.  You've been running the archiver job to archive incoming log messages to S3.
2.  DISASTER!  A bad curator run deletes the last 2 weeks indexes; including todays.
3.  In order of priority, users care about:
   1.  Data from the last 5 minutes - which is currently being shipped in
   2.  Data from the last hour, which needs to be restored
   3.  Data from the last day, which needs to be restored
   4.  Data from the last week
   5.  Data from the 2 weeks ago.

Could you walk through how the archiver restorer job would be used.

Thx

dpb587:
This only contains the plumbing for the task. The actual restoration process is dependent on various factors and gets more complex if you need additional fault tolerance (e.g. if restorer job gets terminated in the middle of everything) and wasn't something I planned on committing here. The process would look like:
1. Add a restorer job to the deployment. For fault tolerance, you'd want to run it on a persistent disk and configure `ingestor_archiver.data_dir` to it. If the instance gets terminated, it'll restart restoring the logs from where it stopped. Optionally add additional restorer jobs (or logstash parsers pointed to the restorer queue) for faster importing.
2. Find all the archived files you'll want to restore. You'll most likely want to sort them according to recency.
3. Push those files up to the restorer job so they're local files. With S3 you'll either need to remotely run authenticated `wget`s or do locally download them and then upload them to restorer; or other more complicated methods.
4. Run the `reload` script on restorer, passing the name of the local file.
5. In a batched environment, you'll probably want to use the `wait-until-loaded` script to block until it's been fully read because logstash won't exit at EOF.
6. Repeat the reload/wait steps until you've restored everything you want.
7. Remove the restorer job(s) from the deployment.

With your specific example, the simplest example is [gisted here](https://gist.github.com/dpb587/1faa9208cfe1adeb99c2) which uses awscli to enumerate archives and download/upload/reload/wait in the grouped ordering. This is obviously not fault tolerant and cannot automatically start where it left off if internet connectivity drops or the restorer instance dies. That requires additional magic and queueing and retrying, and time.

mrdavidlaing:
In the interests of simplicity, lets ignore the fault tolerance aspect for now.

I'm confused as to why we don't just `bosh ssh` into the `restorer` and do all the work from there (possibly using helper scripts that ship with the `restorer` job):
1.  Would it not be simpler (and have less chance of failure) if all fetching from S3 was done on the `restorer` job?  Given that we're probably talking multi GB of archives; that seems that downloading to a location outside AWS would just slow things down dramatically and increase the chance of failure.
2.  You've already given the archiver job AWS credentials; could the `restorer` not just reuse those to fetch the archives?

dpb587:
1. You can do all the work from restorer, if you prefer. The difficult part is enumerating your archive list and getting private files.
2. Restoring requires multiple AWS operations (at least get-object, you probably also want list-objects), which requires additional URL signing. That requires either creating some more BASH scripts for all those signings (which I consider very fragile and confusing) or embedding the AWS CLI (which adds quite a few dependencies, namely python + awscli + awscli's pip dependencies).

Additional notes which affected my thoughts...

I don't think it's practical to download every archived file that we want to the restorer, extract them, and let it run on those. That's not necessarily scalable, but more importantly it can cause unrecoverable stress on other components of the system. For this reason I did a batching approach.

Ingestion will go much more quickly than parsing. If you're doing some bulk restoring with lots of data (e.g. hundreds of GBs) operators need to be wary of:
- if restoring all of those files at once, the restorer's disks (persistent or not) needs to be very large
- overloading redis, affecting both real-time, archive, and restoration queues.

This can be fixed in multiple ways, depending on what's most practical for them:
- use sufficiently large instances with large ephemeral disks; or configure very large persistent disks
- use a separate redis server dedicated to restoration to avoid accidentally overloading the real-time processing (this is just a safety net, though; separate redis server could still easily get overloaded and crash itself, just it won't affect real-time)
- using a very large number of logstash parsers (however you're then likely to run into elasticsearch bottlenecks)
- run restore tasks in batches and add some naive (e.g. sleep a couple minutes between batches) or intelligent sanity checks (e.g. wait to do the next batch until queue is nearly 0).

The archived files are created on a user-configurable schedule. We defaulted to half hour because it's a convenient balance of timespan and file size for us. Smaller setups will probably do them hourly. This makes it slightly more difficult to create a one-size fits all restoration script.

I took the approach that the restorer job template is a utility. I don't much care how it's used, because different people will want to use it different ways. Personally I think it's simplest to do things remotely like in the examples because I don't like working on BOSH machines due to their inherently limited tooling. I envisioned more advanced people building services on top of logsearch for their own needs would have their own 'sidekick' job templates (e.g. one which adds a service monitoring a dedicated SQS queue for archived file URLs which it downloads, kicks to ingestion, and intelligently waits for 0 queue and local reading to complete before continuing). But, those sorts of additional and complex jobs don't seem nearly worth it right now because they add further dependency on AWS/S3 and, namely, we haven't been restoring archived files to know what will actually be the most useful approach for us.

PR #85, opened 2014-11-04
Switch parsers to using a local elasticsearch client node

dpb587:
The logstash parsers currently have a single point of failure because they can only talk to one elasticsearch host at a time. The alternatives are to switch to logstash's embedded elasticsearch client or to run your own in order to utilize elasticsearch's built-in cluster discovery capabilities.

Here I've opted to using our own packaged version rather than the (generally outdated) embedded version logstash has because it allows us to run the same elasticsearch version throughout the cluster and avoid the occasional bugs in running differing versions. This also lets us continue to use the sonian plugin to log the bulk indexing requests logstash makes and graph those times. Since parsers should be very ephemeral, elasticsearch is configured as a client node and will not become master nor will it store any data.

I've included a boolean property `logstash_parser.use_local_elasticsearch` which allows disabling the functionality (in the case of co-locating everything on micro installs).

### DIFF
--- a/jobs/log_parser/monit
+++ b/jobs/log_parser/monit
@@ -3,3 +3,11 @@ check process log_parser
   start program "/var/vcap/jobs/log_parser/bin/monit_debugger log_parser_ctl '/var/vcap/jobs/log_parser/bin/log_parser_ctl start'"
   stop program "/var/vcap/jobs/log_parser/bin/monit_debugger log_parser_ctl '/var/vcap/jobs/log_parser/bin/log_parser_ctl stop'"
   group vcap
+
+<% if p('logstash_parser.use_local_elasticsearch') %>
+  check process log_parser-elasticsearch
+    with pidfile /var/vcap/sys/run/log_parser/elasticsearch.pid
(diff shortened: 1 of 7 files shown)


dpb587:
Waiting on merging to first deploy and validate this in our clusters...

dpb587:
We successfully deployed this to our clusters with the following results on the larger one...

Given the following as the baseline for throughput...

screen shot 2014-11-06 at 2 49 06 pm

CPU utilization on log parsers seems to have gone up slightly with the addition of elasticsearch on the nodes.

screen shot 2014-11-06 at 2 35 11 pm

Network Output on log parsers went down dramatically, presumably due to elasticsearch transport being so much more efficient than HTTP...

screen shot 2014-11-06 at 2 38 39 pm

API frontend node - with parsers shipping directly into elasticsearch, this obviously went down dramatically...

screen shot 2014-11-06 at 2 51 20 pm

These changes have reduced the API SPOF and made it and our parsing a bit more scalable.

Merging and closing...

PR #94, opened 2014-12-17
Added scp support to the archiver job.

malston:
This PR provides another method for archiving your logs. In particular,
when you don't have an outside connection from the network to the
internet, you may want to archive your jobs to another internal location.
It creates a ssh keypair for passwordless ssh and uses the sshpass library
in conjuction with ssh-copy to distribute the public key to the archive machine
without user interaction. Once the key is transferred, the public key is
removed from the archiver job vm.


dpb587:
Thanks for sending this PR - I'm very excited to see it. A couple comments and questions...

Isn't the `distr_sshkey` script being called every time cron executes it, and is this intentional? To me, this seems like it should be, at most, a one-time event, per job. The only reason I can think of needing this run every time is if people have some crazy DNS/instance rotation scenario.

I'm not a huge fan of requiring plain passwords like this - I think SSH keys are more typical in enterprise environments. Is password usage a requirement you have in your environment for some reason? To me it makes more sense to always use a single form of authentication (SSH keys) and keep the scripts/package requirements much simpler. Keys have the added benefit of being externally managed by whatever is responsible for the destination server (e.g. puppet or bosh job) and additional security benefits, such as only allowing specific commands to be run (e.g. only allowing a key to use sftp to drop off files in a specific path). What would you think about converting `archiver.scp.password` to be `archiver.scp.ssh_key`, with the value being written to a file and referenced like `scp -i /var/vcap/jobs/archiver/config/scp.pem ...`?

malston:
1. You're right. distr_sshkey should only be called once. Any suggestions on how to implement that?
2. I like the idea of not ever requiring a password. In this implementation I require the password for the "one time" transfer of the ssh key that is generated by the job. However, I like your approach much better.

dpb587:
For doing `distr_sshkey` stuff once, I'd use `ssh-keygen -f custompath.pem ...` and write it to a specific location. Then you could wrap the key-creation and `ssh-copy-id` in a conditional checking that the file doesn't already exist. But if switching from a password to an SSH key as a job property, the `distr_sshkey` steps will no longer be needed.

dpb587:
@malston, for the longest time I thought this was still waiting for a commit switching from semi-interactive to fully key-based as we had discussed. Looking at it today though, it looks like you took care of that a long time ago and I think this can/should be merged. Any objection to me still going ahead and merging this?

malston:
None by me.

dpb587:
Merged - thanks!

Issue #93, opened 2014-12-03
elasticsearch: Too many open files


dpb587:
I'm getting a `Too many open files` error from elasticsearch on one of my clusters...

```
java.nio.file.FileSystemException: /var/vcap/store/elasticsearch/logsearch/nodes/0/indices/logstash-2014.11.30/3/index/_27h_Lucene410_0.dvm: Too many open files
```

The active ulimit is `4096`...

```
$ cat /proc/$( cat /var/vcap/sys/run/elasticsearch/elasticsearch.pid )/limits | grep files
Max open files            4096                 4096                 files     
```

It's using more than 100 file descriptors on a single index...

```
$ ls -l /proc/$( cat /var/vcap/sys/run/elasticsearch/elasticsearch.pid )/fd  | grep logstash-2014.11.30 | wc -l
104
```

I was able to temporarily fix it by closing some of the older indices.

The node was running with about 37 indices, 149 shards on an instance with 8 cores, 30GB RAM.

We should probably be setting ulimit to the [recommended settings](http://www.elasticsearch.org/tutorials/too-many-open-files/) like we did [previously](https://github.com/cityindex/logsearch/blob/85392e240a960a2ee936a7171901d49b1a0b4ebd/Rakefile#L41). I'm surprised we haven't run into this issue on our larger clusters.

dpb587:
Resolved by #101.

PR #101, opened 2015-02-06
Set ulimit -n 64000 on elasticsearch

dpb587:
Issue #93

### DIFF
--- a/jobs/elasticsearch/templates/bin/elasticsearch_ctl
+++ b/jobs/elasticsearch/templates/bin/elasticsearch_ctl
@@ -54,6 +54,8 @@ case $1 in
 
     /var/vcap/jobs/elasticsearch/bin/undrain 2>&1 >> $LOG_DIR/undrain.log &
 
+    ulimit -n 64000
+
     exec /var/vcap/packages/elasticsearch/bin/elasticsearch \
          -Des.config=${JOB_DIR}/config/config.yml \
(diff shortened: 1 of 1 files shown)


dpb587:
@mrdavidlaing, is `ulimit` "bosh" friendly? Or is there a more cross-OS friendly way that you've seen releases use?

dpb587:
Also, we really should be running all our jobs under the `vcap` user; but that should probably be a separate issue.

mrdavidlaing:
@dpb587 - `ulimit` is the way to go.  BOSH is pretty much confined to running on Ubuntu & CentOS; I wouldn't worry about trying to make logsearch-boshrelease work across anything more than those (and actually, I'd only worry about Ubuntu unless someone complains)

dpb587:
Deployed to meta. Before and after...

```
$ wget -qO- api.meta.logsearch.io:9200/_nodes/_all/process | jq '.nodes | map(select("false" != .attributes.data)) | map({ "key" : .name , "value" : .process.max_file_descriptors }) | sort_by(.key) | from_entries'
{
  "elasticsearch_persistent-z1/0": 4096,
  "elasticsearch_persistent/0": 4096,
}
$ !!
{
  "elasticsearch_persistent-z1/0": 64000,
  "elasticsearch_persistent/0": 64000
}
```

PR #103, opened 2015-02-12
correcting logic for make_manifest script

haydonryan:
The logic was broken in the make manifest script.

PR #104, opened 2015-02-15
Better sample manifests

mrdavidlaing:
This PR removes the unused `spiff` based templates, and replaces them with a collection of sample manifests for different clouds / cluster sizes.

It's related to #103


dpb587:
I consider internal/external IPs, security groups, subnets, S3 buckets/paths, etc to all be internal identifiers which should never be unnecessarily published. I think they should be scrubbed and normalized into test values before rebasing and merging.

drnic:
For anyone looking, https://github.com/cloudfoundry-community/logsearch-boshworkspace has some spiff templates (and uses bosh-workspace to compose them together)

dpb587:
@mrdavidlaing, should this be merged? I commented that I disliked our internal identifiers being published (despite not necessarily confidential), but I'd rather have this merged and closed than hanging around indefinitely.

mrdavidlaing:
This PR has served its purpose of getting something out quickly; I'll close it since its now longer relevant now that projects like https://github.com/cloudfoundry-community/logsearch-boshworkspace exist.

Issue #68, opened 2014-07-21
Customizable logstash index format


dpb587:
By default, logstash uses the index format of `logstash-%{+YYYY.MM.dd}`, creating a single index per day. However, I think there's a reasonable case for wanting to be able to configure it, depending on the installation requirements:
- for smaller or significantly larger logging requirements, it can be more efficient for both kibana usage and elasticsearch to use a different index date format (e.g. `logstash-%{+YYYY-ww}` for weekly)
- instead of using separate deployments, it can be more effective to segment indices by some property (e.g. `logstash-%{@environment}-%{+YYYY.MM.dd}` for environment-specific indices)

We could expose this as a `logstash_parser.index_format` property, defaulting to the current setting of `logstash-%{+YYYY.MM.dd}`.

dpb587:
@mrdavidlaing, I don't think this is a requirement within the CI scope, so, if I sent a PR from personal time for this, would you consider it a worthwhile feature to merge?

mrdavidlaing:
Of course.

I'm interested in your analysis of the circumstances when it makes sense to have different index segmentation.

In the CI use case, access to the last hour of data is more important than access to earlier data, so if changing the index segmentation can assist with that, then its perhaps a feature we can use there too.

dpb587:
Resolved by #106.

PR #106, opened 2015-02-26
Allow customizing document id, index, and index type

dpb587:
This allows users to customize the `document_id`, `index`, and `index_type` options of the `elasticsearch_http` output used by the log parsers. The `index` and `index_type` are the main ones I'm interested in (to support weekly indices and multi-field type names). I went ahead and included `document_id` since `l4aws` was overwriting it and it kind of correlates with index/type name settings, but I don't really have a specific need for it at the moment.

The three new properties are all defaulted to what the existing values have been.

Partially in reference to issue #68.

### DIFF
--- a/jobs/log_parser/spec
+++ b/jobs/log_parser/spec
@@ -40,6 +40,15 @@ properties:
     default: auto
   logstash_parser.idle_flush_time:
     description: "How frequently to flush events if the output queue is not full."
+  logstash_parser.elasticsearch_document_id:
+    description: "Use a specific, dynamic ID rather than an auto-generated identifier."
+    default: ~
+  logstash_parser.elasticsearch_index:
+    description: "The specific, dynamic index name to write events to."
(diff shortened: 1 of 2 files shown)


dpb587:
@mrdavidlaing, any objections to merging?

mrdavidlaing:
@dpb587 - looks good to me.  Merge it.

Out of interest, does yaml like this:

```
default: ~
```

mean  `default := null` ?

dpb587:
I don't recognize the syntax `default := null`, so I'm not sure what you're asking. The `default` is a BOSH key, not YAML; and I'm just setting the property value to [`null`](http://yaml.org/type/null.html) so users don't need to set it in their manifests.

mrdavidlaing:
@dpb587 I was just wondering what `default: ~` means.

`default:=null` would be how you would assign the variable `default` to `null` in Pascal (or Golang)

PR #108, opened 2015-03-13
Update to latest ELK stack

mrdavidlaing:
The PR upgrades all the ELK components to:
1.  Java8
2.  Elasticsearch 1.4.4
3.  Logstash 1.5.0.rc2
4.  Kibana 4.0.1


packages/logstash/packaging
```diff
@@ -6,5 +6,16 @@ set -u # report the usage of uninitialized variables
...
+JAVA_HOME=/var/vcap/packages/java8
+PATH=$PATH:/var/vcap/packages/java8/bin
+
+# Install plugins
+
+# After https://github.com/elastic/logstash/issues/2674 gets resolved, install plugins from .gem files like this:
+# $BOSH_INSTALL_TARGET/logstash/bin/plugin install $BOSH_COMPILE_TARGET/logstash/logstash-filter-alter-0.1.5.gem
+
+# For now, we have to install from rubygems; meaning that the compilation VMs must have 'net access
+$BOSH_INSTALL_TARGET/logstash/bin/plugin install logstash-filter-alter
```
mrdavidlaing: @dpb587 - Any ideas on what to do w.r.t which logstash 1.5.0 plugins to install?

What do you think to just installing them all?

dpb587: After the referenced bug is fixed. when installing .gem files, is an internet connection required? What about for more complex ones which may have a dependency on another third-party gem?

dpb587:
You're brave upgrading both software and its runtime in a single step. Personally I would have been a bit more conservative and done java after we successfully deployed and confirmed the other upgrades. But, if you're confident everything is working, I'm good with that.

I didn't realize logstash 1.5 hadn't been release yet - are we going to wait to merge until 1.5 is stable and we've updated this branch with it?

packages/logstash/packaging
mrdavidlaing: > when installing .gem files, is an internet connection required

In theory no.

However, 

> What about for more complex ones

I don't know :disappointed:

mrdavidlaing:
It turns out that this upgrade is going to be breaking; so I'm inclined to bundle all the breaks into one release.

Specifically:
- Logstash 1.5 breaks some filters (specificaly those that muck about with ruby `Time`)
- Java8 disables SSLv3, which breaks nxlog < 2.9

mrdavidlaing:
> are we going to wait to merge until 1.5 is stable and we've updated this branch with it

Lets wait up to 2 weeks, and then make this call again.

dpb587:
logsearch-shipper then also needs to be updated and redeployed to all the deployments using it - otherwise it sounds like encrypted syslog forwarding is going to break when the target logsearch is upgraded.

mrdavidlaing:
Yes - logsearch-shipper needs to be upgraded

mrdavidlaing:
@dpb587, You can now enable SSLv3 support (it's disabled by default) by adding this to your deploy manifest:

```
logstash_ingestor:
    syslog_tls:
        jdk_tls_client_protocols:"SSLv3,TLSv1,TLSv1.1,TLSv1.2"
```

`logsearch-shipper` should still be upgraded to nxlog 2.9, but it isn't required to deploy logsearch v19

dpb587:
Great, thanks.

mrdavidlaing:
@dpb587, I've pulled `logsearch/logsearch-filters-common` into `src/logsearch-filters-common` as we discussed.

After removing (see https://github.com/cityindex/logsearch-filters-internal/pull/55) the `syslog` filters from `cityindex/logsearch-filters-internal`, the resulting running `/var/vcap/jobs/log_parser/config/logstash.conf` is the same, as expected.

I'm fairly pleased with the outcome.  Any further thoughts?

mrdavidlaing:
BUG:  It turns out that for SSLv3 to be supported the `/var/vcap/packages/java8/lib/security/java.security` file must be edited and `SSLv3` removed from the `jdk.tls.disabledAlgorithms=SSLv3` setting

jobs/kibana/spec
```diff
@@ -11,9 +10,9 @@ templates:
   helpers/ctl_setup.sh: "helpers/ctl_setup.sh"
   helpers/ctl_utils.sh: "helpers/ctl_utils.sh"
 properties:
-  kibana.elasticsearch:
-    description: "IP address and port of elasticsearch host to proxy requests for (eg, 127.0.0.1:9200)"
-    default: "http://127.0.0.1:9200"
+  kibana.api_host:
```
dpb587: Can we keep this as elasticsearch? Since kibana doesn't need to know the difference on whether it's logsearch's api or elasticsearch's api, it seems better to stick with expected vendor conventions.

jobs/kibana/templates/bin/kibana_ctl
```diff
@@ -5,14 +5,22 @@ set -u # report the usage of uninitialized variables
...
-    pid_guard $PIDFILE kibana
+    
+    # Wait up to 5 min for the kibana.api_host to be ready to prevent a startup failure
+    n=0
+    until [ $n -ge 60 ]
+    do
+      curl -s <%= p('kibana.api_host') %> | grep -q 200 && break
+      n=$[$n+1]
+      sleep 5s
+    done
```
dpb587, line 21: Can you explain to me why this whole block is necessary?

jobs/kibana/templates/config/kibana.conf.erb
```diff
@@ -5,35 +5,56 @@ port: <%= p('kibana.port') %>
...
+
 # The default application to load.
 default_app_id: "<%= p('kibana.default_app_id') %>"
 
-# Time in seconds to wait for responses from the back end or elasticsearch.
-# Note this should always be higher than "shard_timeout".
+# Time in milliseconds to wait for responses from the back end or elasticsearch.
 # This must be > 0
-request_timeout: 60
+request_timeout: 300000
```
dpb587: These shard and request timeouts are heavily data and cluster dependent. I think they should be configurable (and defaulted to kibana's built-in defaults, if they're not already).

jobs/kibana/templates/bin/kibana_ctl
```diff
@@ -5,14 +5,22 @@ set -u # report the usage of uninitialized variables
...
 
-    echo $$ > $PIDFILE
+    pid_guard $PIDFILE kibana
```
dpb587, line 23: This is using exec - if `/var/vcap/packages/kibana/bin/kibana` is forking, the `exec` term should be removed; otherwise, if it's not forking and as the parent process `$PIDFILE` should remain `$$`.

jobs/log_parser/templates/bin/log_parser_ctl
```diff
@@ -26,11 +26,10 @@ case $1 in
     # construct a complete config file from all the fragments
     cat ${JOB_DIR}/config/input_redis_and_output_elasticsearch.conf > ${JOB_DIR}/config/logstash.conf
     cat ${JOB_DIR}/config/filters_pre.conf >> ${JOB_DIR}/config/logstash.conf
+    cat ${JOB_DIR}/config/logstash-filters-common.conf >> ${JOB_DIR}/config/logstash.conf
 
     <% if '' != p('logstash_parser.filters') %>
       cat ${JOB_DIR}/config/filters_override.conf >> ${JOB_DIR}/config/logstash.conf
-    <% else %>
-      cat ${JOB_DIR}/config/filters_default.conf >> ${JOB_DIR}/config/logstash.conf
```
dpb587, line 33: Note: this means we're no longer shipping some default filters that the experimental user can play with which I think we may be mentioned in the docs somewhere (which I'm perfectly okay with and support, but not something we previously discussed).

src/logsearch-filters-common/Rakefile
```diff
@@ -0,0 +1,37 @@
+require 'erb'
```
dpb587, line 1: This file has been unused since we have `bin` scripts. It should be deleted.

dpb587, line 1: Apparently I was wrong. Nevermind.

src/logsearch-filters-common/target/logstash-filters-default.conf.erb
```diff
@@ -0,0 +1,128 @@
+if [@type] in ["syslog", "relp"] {
```
dpb587, line 1: I'm not a huge fan of having two compiled files committed. On the one hand, this is a static file, so it's much more appropriate to be included in the logstash package, bypassing the ERB double-escaping rationale for this particular file. On the other hand, I do feel like we might have a need to make something here user-configurable (e.g. default timezone).

jobs/kibana/templates/bin/kibana_ctl
dpb587, line 21: Answer: kibana start up will fail if elasticsearch isn't available

jobs/kibana/spec
mrdavidlaing: OK

jobs/kibana/templates/config/kibana.conf.erb
mrdavidlaing: OK

jobs/log_parser/templates/bin/log_parser_ctl
mrdavidlaing, line 33: Agreed.  In logsearch v19+, you get bundled filters which do "transport" related things, like parse syslog; but nothing further.

src/logsearch-filters-common/target/logstash-filters-default.conf.erb
mrdavidlaing, line 1: After agreeing that this should be moved into the logstash package, I actually think that your final argument `we might have a need to make something here user-configurable` is actually the clincher as to why it should remain in the job.  (which in turn requires commiting 2 copies of the compiled file; one which "normal", and one which is "escaped" so it can be passed through BOSH's template generating ERB process)

mrdavidlaing:
@dpb587, could you peer review these changes please

mrdavidlaing:
@dpb587 I've deployed this to meta.logsearch.io - https://github.com/cityindex/labs-operations/compare/2135b58017d5...9719435c4add - and it seems to be working ok.

mrdavidlaing:
@dpb587, Its also deployed to cityindex.logsearch.io - https://github.com/cityindex/labs-operations/compare/9719435c4add...cbfe776225d4

If you are happy after review, please just go ahead and merge this PR.

PR #117, opened 2015-04-15
Adding skip option

johnnydtan:
For the rolling restart script, this adds the ability to skip a node. This is especially handy if the script originally got interrupted and/or some nodes don't need maintenance.


johnnydtan:
It has the side benefit, for those running monitoring or config management agents that restart processes, of allowing one to control when the next node goes into maintenance.

mrdavidlaing:
@johnnydtan,

Firstly, thanks for the PR!

Secondly, before we can merge the PR, I need a signed CLA from you - please see https://github.com/logsearch/licensing/ for details.

Finally, could you modify your change so you can pass a list of nodes to skip as an arg / env variable, rather than making the script interactive. The reason is that on a largish size cluster you wouldn't want to wait around for the multihour process to push enter every 10 - 120 minutes. 

Thanks!

johnnydtan:
Sent.

I modified it so that you have to explicitly request interactive via `-i` or `--interactive`. This way, default behavior is kept.

dpb587:
Merged - thanks!

PR #118, opened 2015-04-28
Adds ability to specify input codec as a manifest property.

bsnchan:
Signed-off-by: Derwei Chan dchan@pivotal.io

This allows us to enable the multiline codec.

### DIFF
--- a/jobs/log_parser/spec
+++ b/jobs/log_parser/spec
@@ -49,6 +49,8 @@ properties:
   logstash_parser.elasticsearch_index_type:
     description: "The specific, dynamic index type name to write events to."
     default: "%{@type}"
+  logstash_parser.input_codec:
+    description: "The input codec and configuration used for input data."
   logstash_parser.use_local_elasticsearch:
     description: "Run a local elasticsearch client node"
     default: true
--- a/jobs/log_parser/templates/config/input_redis_and_output_elasticsearch.conf.erb
+++ b/jobs/log_parser/templates/config/input_redis_and_output_elasticsearch.conf.erb
@@ -7,6 +7,9 @@ input {
         data_type => "list"
         key => "<%= p("redis.key") %>"
         threads => 8
+        <% if p("logstash_parser.input_codec") %>
+        codec => <%= p("logstash_parser.input_codec") %>
+        <% end %>
     }
 }
 


dpb587:
Hi - thanks for the PR! I have a question about your use case, if you don't mind. First some background and our understanding. We've experimented with multiline before (although with the filter, not the codec), but moved it out of the release and our manifest configuration because we felt it didn't scale on the deployment side of things.

In most cases, log parsers are one of the first bottlenecks in the deployment and it gets scaled by adding another job/vm. Once more than one log parser is running, ingested messages will very likely and unpredictably be split across the multiple log parsers without respect to their multiline nature. For example, messages A1 A2 B1 B2 C1 D1 are waiting in the queue - log_parser/0 receives a batch, A1 A1 B1, and log_parser/1 receives a batch, B2 C1 D1. With that, A1+A2 would be joined, but B1 and B2 were sent to separate jobs. Further, since B2 is the second part and would trigger the multiline, it might get appended to Z1, the last message that log_parser/1 received despite having no relation to it.

Even if we were using a single log parser, but had multiple shippers pushing messages in we could still have the same issue. For example, if shipper/0 has A1 A2 B1 B2 and shipper/1 has C1 D1, at some point they will send a batch, even if it's not all the events it has. So, the ingestor could receive/queue the first batch from shipper/0 with A1 A2 B1, receive a batch from shipper/1 with C1 D1, and then receive the next batch from shipper/0 with B2. The log parser would emit A1+A2, B1, C1, D1+B2 since C1 looked like an independent line, but then B2 looked like it needed to be joined.

To avoid this, we moved the multiline configurations to the log shippers. Typically, they know best what kind of logs they're watching and what general conventions the log emitters are using, so we actually found that to be a fairly straightforward process for us. Once we did that we no longer had to worry about maintaining that state once those messages were reaching the logsearch deployment.

So... all that being said, is this PR for the multiline codec the correct solution for your use case? Mainly, I guess, is have you been able to test and verify that this change works for what you need? If so, we can definitely merge it - I just want to make sure we fully understand why it was needed for maintaining it further down the road.

mrdavidlaing:
@bsnchan - any thoughts?

dpb587:
Closing due to inactivity.

@bsnchan, please comment/reopen if we can continue the discussion.

PR #119, opened 2015-04-29
Upgrade to logstash 1.5.0 rc3

mrdavidlaing:
Upgrades logstash from 1.5.0-rc2 to logstash 1.5.0-rc3, which:
- Has performance improvements
- Allows plugins to be installed from local gem files (meaning tha the logstash package compilation should no longer need internet access; and is marginally faster)

### DIFF
--- a/config/blobs.yml
+++ b/config/blobs.yml
@@ -163,3 +163,19 @@ elasticsearch/elasticsearch-1.5.1.tar.gz:
   object_id: 93b93692-d68a-44e7-86d7-dabe5e13f8d5
   sha: b3863a63d265486332042246bf1c002b3a70b46f
   size: 28183803
+logstash/logstash-1.5.0-rc3.tar.gz:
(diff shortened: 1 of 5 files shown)


mrdavidlaing:
@dpb587 - could you review this please.

As best I can tell there are no breaking changes here - the logsearch-for-cityindex test pass with no changes.

dpb587:
Looks fine. I think you should deploy it to meta before merging.

Reminds me we still need to figure out plugin support.

mrdavidlaing:
I'll get this deployed to meta shortly.

Concerning plugin support, I think we copy the way we allow adding plugins to elasticsearch by listing them in your manifest.  I wish there was a way we could install plugins at compile time rather than deploy time; but I can't think of one.

dpb587:
I think this was successfully deployed and safe to merge, right?

PR #122, opened 2015-05-05
Add support for dynamic logstash plugins

dpb587:
So, nearly the same as we handle elasticsearch plugins:

```
properties:
  # values are arrays, first being the installed plugin name, second being install source
  logstash_parser.plugins:
    # one of the built-in, official plugins
    - { "logstash-filter-punct" : "logstash-filter-punct" }
    # an internal plugin
    - { "cityindex-logstash-filter-acctlookup" : "https://s3.amazonaws.com/.../cityindex-logstash-filter-acctlookup-1.2.9.gem" }
```

### DIFF
--- a/jobs/log_parser/spec
+++ b/jobs/log_parser/spec
@@ -53,7 +53,7 @@ properties:
     description: "Run a local elasticsearch client node"
     default: true
   logstash_parser.plugins:
-    description: "Plugins to run logstash with (array[] = { plugin-name: install-source }"
+    description: "Plugins to run logstash with (array[] = { plugin-name: install-source }; e.g. [ { logstash-filter-cityindex-acctlookup: 'https://s3.amazonaws.com/.../logstash-filter-cityindex-acctlookup-1.2.9.gem' } ])"
     default: []
 
   elasticsearch.host:
--- a/jobs/log_parser/spec
+++ b/jobs/log_parser/spec
@@ -52,6 +52,9 @@ properties:
   logstash_parser.use_local_elasticsearch:
     description: "Run a local elasticsearch client node"
     default: true
+  logstash_parser.plugins:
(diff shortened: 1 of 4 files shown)


dpb587:
This was deployed to meta+cityindex.

Review, merge, and create release as ready, @mrdavidlaing.

jobs/log_parser/spec
```diff
@@ -52,6 +52,9 @@ properties:
   logstash_parser.use_local_elasticsearch:
     description: "Run a local elasticsearch client node"
     default: true
+  logstash_parser.plugins:
+    description: "Plugins to run logstash with (array[] = { plugin-name: install-source }"
```
mrdavidlaing: @dpb587 - this description `(array[] = { plugin-name: install-source }` confuses me.  Is it possible to give a more verbose description?, eg:

```
  # values are arrays, first being the installed plugin name, second being install source
  logstash_parser.plugins:
    # one of the built-in, official plugins
    - [ "logstash-filter-punct", "logstash-filter-punct" ]
    # an internal plugin
    - [ "cityindex-logstash-filter-acctlookup", "https://s3.amazonaws.com/.../cityindex-logstash-filter-acctlookup-1.2.9.gem" ]
```

dpb587:
@mrdavidlaing, I added an example to the property description (and to elasticsearch where I originally copied it from). In general, I don't feel like the spec property description is a very good place for documentation so I was trying to keep it concise. It also made me realize I implemented the logstash plugins slightly differently, so I switched to key-value (like elasticsearch) so they're the same.

mrdavidlaing:
That's much clearer to me.

One advantage to the `spec.description:` is that it gets exposed on bosh.io

screen shot 2015-05-06 at 19 15 40

mrdavidlaing:
anywho; I'm happy for you to merge when ready

PR #124, opened 2015-05-07
Basic auth

yuanzhao:
add basic auth for kibana, so that we can avoid some one not related from accessing our log data.


dpb587:
Hi - thanks for breaking these changes into a separate PR! I think it's much easier to discuss that way. Looking through the diff and discussing it more, we have some concerns about merging and adding this sort of authentication to the release:
1. Simply adding basic authentication on port 80 does nothing to prevent direct (and much more destructive) access via elasticsearch port 9200. Because of this, we've been relying on cloud-based firewalls to control access. For example, we allow port 80 access on API/kibana to the general network, but only the management network can access port 9200.
2. A single username/password is usually insufficient for most environments - it becomes a simple stopgap to accidental, cross-deployment access, but little more (in my opinion). Particularly when access is already non-destructive and generally read only, as in the case of this api endpoint.
3. You've added a `server` directive, but we already have a `server` directive for that endpoint. Instead of creating duplicate listeners and potentially ambiguous ordering, please put your auth changes in the existing directive.

Conceptually I'm just concerned about us trying to handle authentication - logsearch isn't in a position to do it well because we aren't ever going to be controlling direct access to all endpoints (e.g. elasticsearch:9200) - it's too complex and varied of a solution. Instead, we'd rather take the position of being the lower-level backend and letting organizations create their own, application-specific frontend tools around this backend. As an example, we've also started a [logsearch-for-cloudfoundry](https://github.com/logsearch/logsearch-for-cloudfoundry) which takes this approach since it integrates CloudFoundry's UAA for authentication to restrict access within its own frontend. Logsearch (core/this repo) obviously can't know everything it needs to about CloudFoundry's authentication or logging strategies, but it can still be the foundation for the platform-specific frontends to do their own release to meet the varied requirements.

If you do need more thorough and fine-grained access controls for your deployment, you may be interested in trying elastic's [Shield](https://www.elastic.co/products/shield) product which deals with that. You can install it on your existing logsearch cluster. If you do happen to go that route, we would be interested to hear about your experience with it and/or see PRs for changes it requires to the API proxy for forwarding and handling credentials like that. Our team has not yet experimented with Shield.

A more minor issue with this PR is that your branch was not correctly merged with our upstream repository, so we wouldn't be able to merge it as is. It seems like you somehow re-committed some of our commits rather than typical merge or rebase methods. Next time, please make sure you've used merge/rebase to start a one-commit PR which is easy for us to review and merge (see https://help.github.com/articles/merging-an-upstream-repository-into-your-fork/ or http://blog.differential.com/best-way-to-merge-a-github-pull-request/ for more details).

It's great having these sorts of tangible pull requests to discuss changes on - I appreciate the time you've taken for that! If you have any more thoughts on all this, we'd be more than happy to continue discussing this.

gaos1:
Hi - Thanks for your comments. We will play with logsearch-for-cloudfoundry and Shield, will create new pull request when necessary.

PR #127, opened 2015-05-18
Remote logsearch-latest.yml

shinji62:
Remove symbolic link as now `bosh upload release` will upload the lastest one.
Btw the link was wrong anyway.

### DIFF
--- a/releases/logsearch-latest.yml
+++ b/releases/logsearch-latest.yml
@@ -1,1 +1,0 @@
-logsearch-.yml


dpb587:
Thanks for the PR and pointing this out! We still have some internal docs and tools referencing the file, so for now I've just gone ahead and [fixed it](https://github.com/logsearch/logsearch-boshrelease/commit/86d551856aa5f3d7885b71c0cfd55fb5987cfa6e) rather than dropping it early.

Issue #116, opened 2015-04-13
Should run processes as vcap


dpb587:
I think we should be running the processes as `vcap`. Some of them (namely elasticsearch) are left running as `root` and that's not great.

We might need to allow logstash ingestors to run as root though since some users might want to bind to lower ports and I don't think logstash is able to drop its privileges once running. Or perhaps make it conditional on the ports they're using, or an optional property flag.

PR #133, opened 2015-06-11
Switch elasticsearch and redis to run as vcap

dpb587:
Which avoids running as root. When upgrading through this, operators would need to correct the file permissions of the following:

```
$ chown -R vcap:vcap /var/vcap/store/elasticsearch
$ chown -R vcap:vcap /var/vcap/store/queue
```

Issue #116


mrdavidlaing:
Lets do the `chown` automatically in:
- [jobs/elasticsearch/templates/bin/elasticsearch_ctl](https://github.com/logsearch/logsearch-boshrelease/blob/run-as-vcap/jobs/elasticsearch/templates/bin/elasticsearch_ctl#L23) and
-  [jobs/log_parser/templates/bin/log_parser_ctl](https://github.com/logsearch/logsearch-boshrelease/blob/run-as-vcap/jobs/log_parser/templates/bin/log_parser_ctl) and 
- [jobs/queue/templates/bin/redis_ct](https://github.com/logsearch/logsearch-boshrelease/blob/run-as-vcap/jobs/queue/templates/bin/redis_ctl)

Having to do it manually will be a real pain on larger clusters and something that is almost guaranteed to catch people out when upgrading with rather dramatic consequences!

Issue #65, opened 2014-07-17
Default @type to unknown


dpb587:
Currently, if an `@type` field is not specified, `@type` becomes `%{@type}` with the interpolation failure. I think it'd be much better to use `unknown` instead.

PR #134, opened 2015-06-11
Avoid indices literally named %{@type}

dpb587:
Abnormal URIs are inconvenient, and `unknown` makes more sense than if you happen to forget to tag a message.

Issue #65


mrdavidlaing:
This seems like a much better default for @type.

:+1: for merging.

PR #135, opened 2015-06-11
Support logstash filters on ingestors for naive filtering and throttling

dpb587:
If people really want throttling, they can configure it with the `logstash_ingestor.filters` property which operates identically to the parsers, just on the ingestors. The usage should obviously be kept to a minimum given parsers are the more scalable option. More intended to be a temporary stopgap to something causing the later services to topple.

Issue #4


mrdavidlaing:
:+1 for merging

PR #136, opened 2015-06-19
use plain text file for logsearch-latest.yml 

gaos1:
With v20, 

```
± |develop ✓| → bosh upload release releases/logsearch-latest.yml

Verifying release...
File exists and readable                                     OK
Extract tarball                                              FAILED
```

It appears that having logsearch-latest.yml as a symlink is the root cause, since 

```
bosh upload release releases/logsearch-20.yml
```

works.

It seems we need to use plain text for logsearch-latest.yml as a symbol link cannot work with bosh. with this change, anyone can always upload release by:

```
bosh upload release releases/logsearch-latest.yml
```

and we should keep updating the file whenever a new release is available.

### DIFF
--- a/releases/logsearch-latest.yml
+++ b/releases/logsearch-latest.yml
@@ -1,1 +1,82 @@
-logsearch-20.yml
+---
+packages:
+- name: elasticsearch
+  version: eca0f2f78fb671a7c22d4baef2542ad68e9b71f8
+  fingerprint: eca0f2f78fb671a7c22d4baef2542ad68e9b71f8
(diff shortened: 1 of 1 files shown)


mrdavidlaing:
@dpb587 I can replicate this issue under Linux.

Since you originally created the symlink, I'm assuming it works for you?  Is that true?

dpb587:
Didn't realize it wasn't working for others.

Merged - thanks!

mrdavidlaing:
:+1:

PR #137, opened 2015-06-19
Add templates and sample stub files for aws and warden

yuanzhao:
This was modelled after [cf-release](https://github.com/cloudfoundry/cf-release/).

### DIFF
--- a/jobs/ingestor_relp/spec
+++ b/jobs/ingestor_relp/spec
@@ -24,6 +24,7 @@ properties:
 
   logstash_ingestor.relp.port: 
     description: Port to listen for RELP messages
+    default: 2514
 
   redis.host: 
     description: Redis host of queue
--- a/templates/logsearch-deployment.yml
+++ b/templates/logsearch-deployment.yml
@@ -1,0 +1,61 @@
+---
+meta:
+  <<: (( merge ))
+  environment: ~
+  default_env:
+    # Default vcap & root password on deployed VMs (ie c1oudc0w)
+    # Generated using mkpasswd -m sha-512
+    bosh:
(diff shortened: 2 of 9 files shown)


templates/logsearch-deployment.yml
```diff
@@ -0,0 +1,61 @@
...
+  max_errors: 1
+
+resource_pools: 
+- name: logsearch
+  network: default
+  stemcell: (( meta.stemcell ))
+  cloud_properties: (( merge ))
+  env: (( merge || meta.default_env ))
+
+- name: logsearch_api
```
dpb587, line 44: Out of curiosity, is there background or a reason for the resource pools all being prefixed with `logsearch`? I thought resource pools were deployment-specific already in which case it seems redundant; but if this sort of thing is good practice I'm interested to learn.

jobs/ingestor_relp/spec
```diff
@@ -24,6 +24,7 @@ properties:
 
   logstash_ingestor.relp.port: 
     description: Port to listen for RELP messages
+    default: 2514
```
dpb587, line 27: Next time, please at least make this sort of change a separate commit. It's very unrelated to the commit message for people reviewing logs to see what changes were happening.

dpb587:
I'm don't really use spiff myself, but I was able to test out the warden setup and these templates will definitely make it more welcoming to others. Thanks for writing them up.

Merged.

templates/logsearch-deployment.yml
jim80net, line 44: No reasoning, though [one day](https://github.com/cloudfoundry/bosh-notes/blob/master/cloud-config.md) BOSH will manage resource pools. Then ours will be to consume, not create. =]

Issue #139, opened 2015-06-22
How we should handle ES plugin?


bonzofenix:
Hi David,
So I been working on getting 2 plugins on ES for our Logsearch deployment:
https://github.com/royrusso/elasticsearch-HQ
https://github.com/lmenezes/elasticsearch-kopf

I got this working by adding 1 package to the release which is called es-plugins
This way I can specified in my properties:

```
...
properties:
    elasticsearch:
    - hq: file:///var/vcap/packages/es-plugins/elasticsearch-hq.zip
    - kopf: file:///var/vcap/packages/es-plugins/elasticsearch-kopf.zip
```

Do you think this worth being merged to the main trunk? 
If thats the case we would need to upload those blobs to the public buckets of yours.

**Pros:**
- Do not need internet connection to use this 2 particular plugins. 
- User can add plugins on src at: src/es-plugins/elasticsearch-PLUGIN_NAME.zip

**Cons:**
- Requires to create a new release whenever you add packages.
  
  What do you think? is this the right approach?

Thanks!

dpb587:
Hi there - thanks for starting a discussion (and PR) about this.

Some background... a while ago, we actually did [include](https://github.com/logsearch/logsearch-boshrelease/commit/6112f23e894c10ab94d2b2128012469a6fd0e7e6) the HQ plugin within the release and used it fairly frequently. After a couple months, we shifted to using the [cloud verison](http://www.elastichq.org/app/index.php) of the plugin simply because they were updating it more frequently than we wanted to keep track of in our release. A few months later, we decided to switch to kopf. A couple months later we had a discussion and decided to [stop](https://github.com/logsearch/logsearch-boshrelease/commit/e51052ce9526552e00dfa2a3ea82058c8a6a261b) including plugins directly in our release. The main rationale was that we didn't want to become responsible (or the bottleneck) for making sure those plugins stayed up to date, that we were including well-known stables ones, handling releases for each upstream bug fix, and worrying about what to do with plugins that we no longer use but others might still prefer to have in their deployments.

Nowadays, we've been recommending the following alternatives which give you much more control about the specific plugins you like and the update schedules you want to follow. The first two will also work for the similar `logstash.plugins` property for when you need some fancier log parsing capabilities.

[...]

PR #140, opened 2015-06-22
Offline ES plugins support. kopf and hq addded.

bonzofenix:
This is the the simplest approach I could find for https://github.com/logsearch/logsearch-boshrelease/issues/139 which would require the user to:
1. add to src the zip of the plugin
2. add the property to install plugin


dpb587:
Closing (discussed further in #139)...

PR #143, opened 2015-07-07
Fixes redis queue logging

bonzofenix:
According to the documentation in the redis config output is being sent to /dev/null.
By configuring this we started getting redis output.


dpb587:
Thanks!

PR #150, opened 2015-07-15
allows merging properties and templates for api job and collocating r…

bonzofenix:
…eleases

This PR allows us to collocate releases into a single manifest. The changes in properties allow us to install plugins into kibana for example.

@malston && @bonzofenix


mrdavidlaing:
@dpb587 - I think this is ready for merging; we have CLAs on file; and I think this is a good extension to the templates

PR #151, opened 2015-07-16
Address e2e errand timing bug

mrdavidlaing:


jobs/test_e2e_errand/templates/bin/run.erb
```diff
@@ -15,40 +16,63 @@ export PATH="/var/vcap/packages/java8/bin:$PATH"
 
 [ ! -e /home/vcap/.sincedb ] || rm /home/vcap/.sincedb
 
-
-echo '' ; echo '==> Purging existing test data...'
-
-curl -s -X DELETE '<%= p('test_e2e_errand.elasticsearch_hostport') %>/_all/sample-logs--repo-commits' ; echo ''
```
dpb587, line 21: I disagree with this being removed. The test should be starting from a known-empty state. Can you explain your reasons?

jobs/test_e2e_errand/templates/bin/run.erb
```diff
@@ -15,40 +16,63 @@ export PATH="/var/vcap/packages/java8/bin:$PATH"
...
+    "filtered" : {
+        "query" : {
+            "term" : { "@source.host":"sample-logs--repo-commits" }
+        },
+        "filter" : {
+          "bool" : {
+            "must" : {
+                "range" : {
+                    "@timestamp" : {
+                        "from" : "$(date --iso-8601=ns)",
```
dpb587, line 30: I'm not sure why `@timestamp` should be necessary here. Without the `DELETE`, there will be multiple records from each test run since the field it is extracted from the commit date (in which case I don't understand how this was working and always returning 32 records). With the `DELETE`, the filter shouldn't be necessary because all records should already be from the specific test's run. Am I missing something?

jobs/test_e2e_errand/templates/bin/run.erb
```diff
@@ -15,40 +16,63 @@ export PATH="/var/vcap/packages/java8/bin:$PATH"
...
+echo '' ; echo '==> Waiting for shipped logs to be ingested and indexed...'
 
-echo '' ; echo '==> Waiting for sanity...'
-
-sleep 5
-
+n=0
+until [ $n -ge 5 ]
+do
+  curl -s -d "$QUERY" -X POST '<%= p('test_e2e_errand.elasticsearch_hostport') %>/_count' > curl.out
```
dpb587, line 56: Originally, I was always being as specific as possible - rather than `/_count`, consider doing `/_all/sample-logs--repo-commits/_count` so you also end up with an implicit test on the type you're expecting.

jobs/test_e2e_errand/templates/bin/run.erb
dpb587, line 30: Just realized it's effectively the test's start date, but still, the `@timestamp` should be the value of the committer date as defined by the [filters](https://github.com/logsearch/logsearch-boshrelease/blob/develop/src/sample-logs/repo-commits/logstash-filters.conf#L21) so I'm still missing how something is working here.

jobs/test_e2e_errand/templates/bin/run.erb
mrdavidlaing, line 21: My understanding is that Delete by Query has been depreciated as of 1.5.3 - https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-delete-by-query.html#docs-delete-by-query
and I can't figure out a straight forward way to replace it.

I also couldn't get this specific delete to work; I think because when I ship the logs they don't get parsed with a `@type:sample-logs--repo-commits`

```
"_source":{"@version":"1","@timestamp":"2015-07-14T17:31:10.427Z","host":"10.12.8.148","@type":"syslog","@message":"<14>1 2015-07-14T17:31:10.427+0000 sample-logs--repo-commits LOGSTASH - - - 9d666a0da317777d35e763800ddfd8507cfb580c b056061ab74cc30ee2e6ea3e530f7262b2245228 2015-05-28 18:36:28 -0600 Danny Berger // Start experimenting with concourse","syslog_pri":["14","14"],"syslog5424_ver":1,"syslog_program":["LOGSTASH","LOGSTASH"],"syslog_message":"9d666a0da317777d35e763800ddfd8507cfb580c b056061ab74cc30ee2e6ea3e530f7262b2245228 2015-05-28 18:36:28 -0600 Danny Berger // Start experimenting with concourse","received_at":["2015-07-14T17:31:10.472Z","2015-07-14T17:31:10.427Z"],"received_from":["10.12.8.148","10.12.8.148"],"tags":["syslog_standard","syslog_standard"],"syslog_severity_code":6,"syslog_facility_code":1,"syslog_facility":"user-level","syslog_severity":"informational","@source.host":"sample-logs--repo-commits","syslog_procid":["-","9d666a0da317777d35e763800ddfd8507cfb580c","-"],"syslog_msgid":["-","b056061ab74cc30ee2e6ea3e530f7262b2245228","-"]}'
```

Which log_parsing rules are you using?

mrdavidlaing, line 21: Aha - from your comment below I see you are expecting https://github.com/logsearch/logsearch-boshrelease/blob/develop/src/sample-logs/repo-commits/logstash-filters.conf to be installed - which I don't currently have

dpb587:
@mrdavidlaing, this PR isn't waiting on anything from me, right?

mrdavidlaing:
@dpb587 - No - its stuck at the bottom of my todo list :(

dpb587:
I'm closing this PR/branch and starting from the latest head because these changes were based on incorrect configuration and this file has since been modified preventing a clean merge.

PR #152, opened 2015-07-16
Added cleanup of test data to end of smoke tests

geofffranks:
There's a beginning step of cleaning up existing data, but this one was added to clean up data post-test, so it isn't sitting around after testing completes until the next deployment/test iteration.


dpb587:
Merged - thanks!

PR #159, opened 2015-07-29
Allows setting es job templates and properties through stub

bonzofenix:
We been working in performing snapshots on ES. For this we mount nfs server on the ES servers. This allow us to collocate templates to the ES jobs in the stub file using spiff.


dpb587:
Merged - sorry about the delay

Issue #73, opened 2014-08-15
Rework queue job to be based on Kafka


mrdavidlaing:
The Logsearch data pipeline currently looks like this:

```
shipper --> ingestor --> queue(redis) --> log_parser --> elasticsearch
```

The queue functions as:
1.  A buffer to prevent spikes in log volume from overloading the datastore
2.  A buffer to enable the datastore to go offline without loosing ingested data
3.  A source of pipeline throughput metrics

Currently the queue runs on a single node, so is a SPOF, and cannot be upgraded without downtime.

[Kafka](http://kafka.apache.org/) whose design goals more closely match the queue job's role.  

Specifically, it is:
- Distributed (by design)
- Designed to buffer several days worth of data, allowing multiple consumers to read data at a speed that suits them.  It's disk based (rather than memory based); and queues can be spread over many machines, so it can buffer a _lot_ more data.
- Fast (because it optimises for the fantastic sequential read performance of modern disks)

I propose that we switch the queue job to be based on Kafka, initially to remove the SPOF, but ultimately to enable our data pipeline to evolve to something like this:

```
shipper --> ingestor --> queue.raw* --> log_parser --> pubsub.decorated*
  --> pubsub.decorated* ----> queue.search* --> elasticsearch
                        |---> queue.longtermstorage* --> S3
                        |---> anomaly detection (reimann)  --> pubsub.notification* ---> elasticsearch
                        |---> anomaly detection (storm)  ---^                      |--> email

* - implemented with Kafka
```

sopel:
:information_source: This details several preceding discussions we had around the subject of integrating [Apache Kafka](https://kafka.apache.org/), which might provide more details regarding the subject matter still accordingly:
- https://github.com/cityindex/logsearch/issues/273 - _Evaluate Apache Kafka_
- https://github.com/cityindex/logsearch/issues/269#issuecomment-30520629 - _Analyze/Implement Auto Scaling for managed queue (Redis)_
- https://github.com/cityindex/logsearch/issues/335 - _Evaluate migration to event sourcing for zero downtime blue green deployments_
- https://github.com/cityindex/logsearch/issues/321 - _Replace EC2 image/snapshot based with Elasticsearch snapshot based backup_

@mrdavidlaing - given I suggested to look into Kafka as the likely ideal solution for several of our long-term scaling and reliability issues, I'm more than happy to back this proposal now :)

mrdavidlaing:
[...]

PR #160, opened 2015-07-30
Added RabbitMQ as input

axelaris:
Hi!
I've just added a RabbitMQ support in addition to Redis.
Would do you like to add it to upstream?

### DIFF
--- a/jobs/parser/spec
+++ b/jobs/parser/spec
@@ -35,7 +35,7 @@ properties:
     default: ''
   logstash_parser.inputs:
     description: |
-      A list of output plugins, with a hash of options for each of them. Please refer to example below.
+      A list of input plugins, with a hash of options for each of them. Please refer to example below.
     example:
       inputs:
         - plugin: rabbitmq
--- a/jobs/parser/spec
+++ b/jobs/parser/spec
@@ -34,7 +34,16 @@ properties:
     description: "The configuration to embed into the logstash filters section"
     default: ''
   logstash_parser.inputs:
-    description: "The configuration to embed into the logstash inputs section"
+    description: |
+      A list of output plugins, with a hash of options for each of them. Please refer to example below.
+    example:
+      inputs:
+        - plugin: rabbitmq
+          options:
+            host: 192.168.1.1
+            user: logsearch
+            password: c1oudbunny
+    default: [ { plugin: "redis", options : {} } ]
   logstash_parser.outputs:
     description: "The configuration to embed into the logstash outputs section"
   logstash_parser.workers:
--- a/jobs/parser/spec
+++ b/jobs/parser/spec
@@ -33,6 +33,8 @@ properties:
   logstash_parser.filters:
     description: "The configuration to embed into the logstash filters section"
     default: ''
+  logstash_parser.inputs:
+    description: "The configuration to embed into the logstash inputs section"
   logstash_parser.outputs:
     description: "The configuration to embed into the logstash outputs section"
   logstash_parser.workers:
@@ -89,13 +91,3 @@ properties:
   redis.key: 
     description: Name of queue to pull messages from
     default: logstash
-
-  rabbitmq.host:
-    description: RabbitMQ host of queue
-  rabbitmq.exchange:
-    description: RabbitMQ Exchange
-    default: amq.direct
-  rabbitmq.user:
-    description: RabbitMQ username
-  rabbitmq.password:
-    description: RabbitMQ password
--- a/jobs/parser/spec
+++ b/jobs/parser/spec
@@ -13,7 +13,7 @@ templates:
   helpers/ctl_utils.sh: helpers/ctl_utils.sh
   config/elasticsearch/config.yml.erb: config/elasticsearch/config.yml
   config/elasticsearch/logging.yml.erb: config/elasticsearch/logging.yml
-  config/input_redis_and_output_elasticsearch.conf.erb: config/input_redis_and_output_elasticsearch.conf
+  config/input_and_output.conf.erb: config/input_and_output.conf
   config/filters_pre.conf.erb: config/filters_pre.conf
   config/filters_post.conf.erb: config/filters_post.conf
   config/logstash-filters-common.conf.erb: config/logstash-filters-common.conf
@@ -89,3 +89,13 @@ properties:
   redis.key: 
     description: Name of queue to pull messages from
     default: logstash
+
+  rabbitmq.host:
+    description: RabbitMQ host of queue
+  rabbitmq.exchange:
+    description: RabbitMQ Exchange
+    default: amq.direct
+  rabbitmq.user:
+    description: RabbitMQ username
+  rabbitmq.password:
+    description: RabbitMQ password
--- a/jobs/parser/templates/bin/parser_ctl
+++ b/jobs/parser/templates/bin/parser_ctl
@@ -49,7 +49,7 @@ case $1 in
     done
 
     # construct a complete config file from all the fragments
-    cat ${JOB_DIR}/config/input_redis_and_output_elasticsearch.conf > ${JOB_DIR}/config/logstash.conf
+    cat ${JOB_DIR}/config/input_and_output.conf > ${JOB_DIR}/config/logstash.conf
     cat ${JOB_DIR}/config/filters_pre.conf >> ${JOB_DIR}/config/logstash.conf
     cat ${JOB_DIR}/config/logstash-filters-common.conf >> ${JOB_DIR}/config/logstash.conf
 
--- a/jobs/parser/templates/config/input_and_output.conf.erb
+++ b/jobs/parser/templates/config/input_and_output.conf.erb
@@ -10,16 +10,11 @@ input {
     }
     <% end %>
 
-   <% if_p("rabbitmq.host") do |rabbitmq_host| %>
-   rabbitmq {
-	host => "<%= p("rabbitmq.host") %>"
-	exchange => "<%= p("rabbitmq.exchange") %>"
-	user => "<%= p("rabbitmq.user") %>"
-	password => "<%= p("rabbitmq.password") %>"
-   }
+   <% if_p("logstash_parser.inputs") do |logstash_parser_inputs| %>
+        <%= p("logstash_parser.inputs") %>
    <% end %>
 
-   <% an_input = p(['redis.host', 'rabbitmq.host'], nil) %>
+   <% an_input = p(['redis.host', 'logstash_parser.inputs'], nil) %>
    <% raise "There are no input configured" if an_input.nil? %>
 }
 
--- a/jobs/parser/templates/config/input_and_output.conf.erb
+++ b/jobs/parser/templates/config/input_and_output.conf.erb
@@ -1,5 +1,5 @@
 input {
-
+    <% if_p("redis.host") do |redis_host| %>
     redis {
         host => "<%= p("redis.host") %>"
         port => "<%= p("redis.port") %>"
@@ -8,6 +8,19 @@ input {
         key => "<%= p("redis.key") %>"
         threads => 8
     }
+    <% end %>
+
+   <% if_p("rabbitmq.host") do |rabbitmq_host| %>
+   rabbitmq {
+	host => "<%= p("rabbitmq.host") %>"
+	exchange => "<%= p("rabbitmq.exchange") %>"
+	user => "<%= p("rabbitmq.user") %>"
+	password => "<%= p("rabbitmq.password") %>"
+   }
+   <% end %>
+
+   <% an_input = p(['redis.host', 'rabbitmq.host'], nil) %>
+   <% raise "There are no input configured" if an_input.nil? %>
 }
 
 output {
--- a/jobs/parser/templates/config/input_and_output.conf.erb
+++ b/jobs/parser/templates/config/input_and_output.conf.erb
@@ -14,6 +14,14 @@ input {
         <%= p("logstash_parser.inputs") %>
    <% end %>
 
+<% p('logstash_parser.inputs').each do | input | %>
+    <%= input['plugin'] %> {
+        <% input['options'].each do | k, v | %>
+          <%= k %> => <%= v.inspect %>
+        <% end %>
+    }
+<% end %>
+
    <% an_input = p(['redis.host', 'logstash_parser.inputs'], nil) %>
    <% raise "There are no input configured" if an_input.nil? %>
 }
--- a/jobs/parser/templates/config/input_and_output.conf.erb
+++ b/jobs/parser/templates/config/input_and_output.conf.erb
@@ -1,25 +1,23 @@
 input {
-    <% if_p("redis.host") do |redis_host| %>
-    redis {
-        host => "<%= p("redis.host") %>"
-        port => "<%= p("redis.port") %>"
-        type => "redis-input"
-        data_type => "list"
-        key => "<%= p("redis.key") %>"
-        threads => 8
-    }
-    <% end %>
 
 <% p('logstash_parser.inputs').each do | input | %>
     <%= input['plugin'] %> {
-        <% input['options'].each do | k, v | %>
-          <%= k %> => <%= v.inspect %>
-        <% end %>
+      <% if 'redis' == input['plugin'] %>
+        <% input['options'] = { 
+           "host" => p('redis.host'), 
+           "port" => p('redis.port'), 
+           "type" => "redis-input", 
+           "data_type" => "list", 
+           "key" => p('redis.key'), 
+           "threads" => 8
+         } %>
+      <% end %>
+      <% input['options'].each do | k, v | %>
+        <%= k %> => <%= v.inspect %>
+      <% end %>
     }
 <% end %>
 
-   <% an_input = p(['redis.host', 'logstash_parser.inputs'], nil) %>
-   <% raise "There are no input configured" if an_input.nil? %>
 }
 
 output {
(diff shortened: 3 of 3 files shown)


mrdavidlaing:
@axelaris Thanks!

Could you give us a little detail about how the RabbitMQ you are using as a queue is configured.  

Out of interest, what makes this better than Redis (for your environment)?

Finally, we'll need a signed CLA from you before we can merge - see https://github.com/logsearch/licensing#signing-the-contributors-license-agreement

axelaris:
Hi David,
please check your mail, I've just mailed signed CLA to you.
As for RabbitMQ configuration, it's a plain server, almost unconfigured. Taken from here: https://github.com/pivotal-cf/cf-rabbitmq-release. I'm using only 'rabbitmq-server' job at the moment.
The reason for swapping Redis with RabbitMQ is HA configuration. My colleagues performed some testing and realised that Redis is quite unstable when using more than one server in stack. Moreover, since I know RabbitMQ has a better performance.
BTW, I've also added the same request for output: https://github.com/logsearch/logsearch-for-cloudfoundry/pull/60

mrdavidlaing:
@axelaris - Thanks for the CLA; everything is good for merging from a legal perspective.

I hear your concerns around Redis as a Single Point Of Failure.  Fixing that has been a long term goal that keeps getting deferred - https://github.com/logsearch/logsearch-boshrelease/issues/73 - because single node Redis has proven extremely stable in production (for us at least).

I'll look to getting this PR merged this week.

Would you consider writing up a blog article for http://www.logsearch.io/blog/ detailing your experience of using Rabbit vs Redis as the queue component?

mrdavidlaing:
@axelaris, I've been discussing this PR with the team; and how it interacts with other "alternate queue" requests we get, eg: Kafka ( https://github.com/logsearch/logsearch-boshrelease/issues/73 ), SQS etc.

We're concerned that going down the route proposed in this PR - where each alternate queue results in more [`queue_type.attribute`](https://github.com/logsearch/logsearch-boshrelease/pull/160/files#diff-41afd21a92eb170d62d62a5ca40ec515R93) spec properties - will lead to configuration confusion and the expectation that we test against multiple queue types (which we don't have the bandwidth to do).

It also means that for every new queue type people want to experiement with we need to update logsearch-boshrelease AND all the related addons that have custom ingestors.

We'd like to propose the following change to your PR to get around this "exploding properties" issue.

Replace the set of `rabbitmq.*` spec properties with a single `logstash_parser.inputs` property that, when set, adds whatever logstash config the operator wants into the `inputs {}` section.  This would work just like the existing [`logstash_parser.outputs`](https://github.com/logsearch/logsearch-boshrelease/pull/160/files#diff-deaee71cabb1bcd6b28fbd8777b0e067R48) property.

Your use case of using RabbitMQ rather than Redis would be supported by:

```
properties:
  logstash_parser:
        inputs: |
                      rabbitmq {
                          host => "10.0.1.x"
                          exchange => "my_rabbit_exchange"
                          user => "logsearch"
                          password => "c1oudbunny"
                      }
```

This way, people wanting to use other queues could use the same `logstash_parser.inputs` to do so.

If you could update this PR as above, we'd be happy to merge.

axelaris:
Hi @mrdavidlaing,
I've just added a commit to swap rabbitmq attributes with logstash_parser.inputs one.

Despite that is does work as expected, I found some disadvantages there. Since input configuration specified in raw format, I can't use variables substitution there (I love to use Spiff much).
Upon deploying RabbitMQ itself, I have to specify the same variables (host, username, password) few times in the manifest. So I believe it was a good idea to specify them once, then use Spiff's substitutions in each other place. But now I can't use Spiff for logstash_parser.inputs.

It would be great to see your thoughts on that.

axelaris:
User will see this exception upon deploying process:

```
ubuntu@zjb:~/elk$ bosh -n deploy
Acting as user 'admin' on deployment 'logsearch' on 'my-bosh'
Getting deployment properties from director...

... skipped ...

Error 100: Error filling in template `input_and_output.conf.erb' for `log_parser/0' (line 18: There are no input configured)

Task 857 error
```

mrdavidlaing:
@axelaris thanks for the update.

I can't think of way to make this work better with spiff; which is unfortunate.  However, I get the feeling that spiffs day's are numbered  - https://github.com/cloudfoundry-incubator/spiff/issues/27#issuecomment-56316106 - ; so maybe whatever arises in its place will give us some better options. (The new [BOSH links feature](https://github.com/cloudfoundry/bosh-notes/blob/master/links.md) looks promising)

Thus, I still think that this is the best tradeoff in allowing people to use alternate queues without having to make changes to the core logsearch boshrelease.

mrdavidlaing:
> User will see this exception upon deploying process:

Cool.  That's really useful.

dpb587:
I have an alternative idea which supports the additional flexibility and possibly avoids the disadvantages mentioned above. What if, instead of `logstash_parser.inputs` being a string, we supported it as a complex array.

The syntax would look something like...

```
logstash_parser:
  inputs:
    - plugin: rabbitmq
      options:
        - host: "10.0.1.x"
        - exchange: "my_rabbit_exchange"
        - user: "logsearch"
        - password: "c1oudbunny"
```

Which would then be ERB like...

```
<% p('logstash_parser.inputs').each do | input | %>
    <%= input['plugin'] %> {
        <% input['options'].each do | option | k, v = option.first %>
          <%= k %> => <%= JSON.generate v.to_s %>
        <% end %>
    }
<% end %>
```

Which would result in a generated template like...

```
rabbitmq {
  host => "10.0.1.x"
  exchange => "my_rabbit_exchange"
  user => "logsearch"
  password => "c1oudbunny"
}
```

Practically, `logstash_parser.inputs[*].options` might be able to be a hash, but since logstash config format supports multiple keys, it seems safer to retain that support here in case `inputs` allow that sort of thing.

I'm unsure whether this sort of complex nesting is compatible with spiff constructs though. @axelaris, since I'm not a spiff user, is this sort of implementation something that would be spiff-friendly and meet your and others' needs?

@mrdavidlaing, is this structured value approach for inputs something we could consider instead?

mrdavidlaing:
@dpb587 - this is an very interesting proposal; at first glance it seems like it might tick all the boxes.

Can we explore how it would work if we wanted to use a `kafka` input ala:

```
kafka {
        zk_connect => "kafka1.example.com:2181"
        topic_id => "logstash"
        reset_beginning => false
        consumer_threads => 1
        decorate_events => true
    }
```

Would the manifest config be:

```
logstash_parser:
  inputs:
    - plugin: kafka
      options:
        - zk_connect : "kafka1.example.com:2181"
        - topic_id: "logstash"
        - reset_beginning: false
        - consumer_threads: 1
        - decorate_events: true
```

@axelaris - I think this would work with `spiff` - can you confirm?

dpb587:
Correct about the kafka example.

axelaris:
First of all, @dpb587, let me thank you for a great idea!
I just tried to apply it and what I got:
With help of `spiff` utility this template:

```
logstash_parser:
    inputs:
      - plugin: rabbitmq
        options:
          - host: (( meta.ip.queue ))
          - exchange: "amq.direct"
          - user: (( properties.rabbitmq-server.administrators.broker.username ))
          - password: (( properties.rabbitmq-server.administrators.broker.password ))
```

has been successfully rendered into this YAML:

```
logstash_parser:
     inputs:
    - options:
      - host: 10.1.0.x
      - exchange: amq.direct
      - user: logsearch
      - password: c1oudbunny
      plugin: rabbitmq
```

But the bad news is that I got an error at deploy time:

```
Error 100: Error filling in template `input_and_output.conf.erb' for `log_parser/0' (line 22: only generation of JSON objects or arrays allowed)
```

Line 22 is a one with `JSON.generate` instruction. Unfortunately, I'm not familiar with Ruby enough to catch the error. Could you?

axelaris:
Fortunately, I was able to make it work by simplifying the ERB code a bit:

```
<% p('logstash_parser.inputs').each do | input | %>
    <%= input['plugin'] %> {
        <% input['options'].each do | k, v | %>
          <%= k %> => <%= v %>
        <% end %>
    }
<% end %>
```

Your thoughts?

mrdavidlaing:
Without JSON.generate, how are none string values rendered, or strings with
a space in them ?

On Wednesday, August 12, 2015, Alexey Zakharov notifications@github.com
wrote:

> Fortunately, I was able to make it work by simplifying the ERB code a bit:
> 
> <% p('logstash_parser.inputs').each do | input | %>
>     <%= input['plugin'] %> {
>         <% input['options'].each do | k, v | %>
>           <%= k %> => <%= v %>
>         <% end %>
>     }
> <% end %>
> 
> Your thoughts?
> 
> —
> Reply to this email directly or view it on GitHub
> https://github.com/logsearch/logsearch-boshrelease/pull/160#issuecomment-130260458
> .

## 

David Laing
logsearch.io - build your own open source cloud logging cluster
http://davidlaing.com

axelaris:
@mrdavidlaing, stings with spaces have been rendered successfully, because of using quotes 
`fake_key: "fake value one"` => `fake_key => fake value one`
As for non-string values, I do not have any real example at the moment. Do you have one?

mrdavidlaing:
@axelaris - I'm pretty sure that logstash config of `fake_key => fake value one` would result in an error.  Should it not be `fake_key => "fake value one"`?

A more complete logstash input config example (that contains all the different "types") that we need to support would be:

```
rabbitmq {
  host => "10.0.1.x"
  exchange => "my_rabbit_exchange"
  user => "logsearch"
  password => "s3cret with space"
  prefetch_count => 512
  durable => true
  tags => ["via_rabbit", "foo", "bar"]
  add_field => ["received_at", "%{@timestamp}", "received_from", "%{@source_host}" ]
}
```

axelaris:
I've checked this out today, and it seems to work really nice!
Here is cut of my stub file:

```
logstash_parser:
    inputs:
      - plugin: rabbitmq
        options:
          host:
          - (( meta.ip.queue ))
          exchange: (( "'" properties.rabbitmq.exchange "'" ))
          user: (( properties.rabbitmq.user ))
          password: (( "'" properties.rabbitmq.password "'" ))
          tags:
          - "via_rabbit"
          - "foo"
          - "bar"
          add_field:
            received_att: "%{@timestamp}"
```

.. and how it's was rendered to Logstash's host:

```
    rabbitmq {
          add_field => {"received_att"=>"%{@timestamp}"}
          exchange => 'amq.direct'
          host => ["10.1.0.x"]
          password => 's3cret with space'
          tags => ["via_rabbit", "foo", "bar"]
          user => logsearch
    }
```

Logstash was started successfully and working smoothly. And I was able to find added tags and field in the Elasticsearch. As you can see, I've added a single quote inside double quotes to pass a quote behind the Spiff. And it's working. 

Just a single thing that I was failed to achieve - add more than one extra field. As was realised, Logstash accepts only one _key=>value_ pair inside **add_field** option. To add another pair we have to use one more **add_field**. But when I add more keys to YAML config - it's will be rendered together in a single array of hashes, which Logstash does not accept. But it's not the point where `JSON.generate` could help.

Any thoughts?

mrdavidlaing:
@axelaris - Looks like we're making great progress!  Thanks!

For `add_field`, have you tried just passing the `key=>values` as an array?  (I think logstash accepts this as an alternate to a hash).  Ie:

```
logstash_parser:
    inputs:
      - plugin: rabbitmq
        options:
          ...snip...
          add_field:
            - "received_at"
            - "%{@timestamp}"
            - "received_from"
            - "%{host}"
```

dpb587:
Regarding escaping issues...

I don't believe we should have to embed quotes into YAML values in order to get things quoted in the logstash config. For example, instead of...

```
options:
  exchange: "'amq.direct'" # becomes...     exchange => 'amq.direct'
```

We should always be practically smart and escape them...

```
options:
  exchange: "amq.direct"   # becomes...    exchange => "amq.direct"
```

I wasn't aware of the `only generation of JSON objects or arrays allowed` message you posted. Instead you should try using the `quirks_mode` option which I think will take care of the escaping issue altogether...

```
<%= k %> => <%= JSON.generate(v, quirks_mode: true) %>
```

---

Regarding `add_field` issues...

Technically, I think @axelaris is correct in using a hash for `add_field` because it is documented as such in their [docs](http://logstash.net/docs/1.4.2/inputs/file#add_field). However, it may also be an array for historical reasons, like we've been [using](https://github.com/logsearch/logsearch-boshrelease/blob/develop/jobs/ingestor_syslog/templates/config/syslog_to_redis.conf.erb#L4).

Note, we have switched away from my original proposal of treating `options` as an array of hashes. With the original proposal, you could do:

```
options:
  - { add_field: [ "name1", "value1" ] } # becomes....    add_field => ["name1","value1"]
  - { add_field: [ "name2", "value2" ] } #                add_field => ["name2","value2"]
```

But, I like the hash-based approach you've been using better because it is simpler and readable. In practice, array is the only config field which can be listed multiple times because it can be internally merged. So, I say continue with the hash like you currently have.

I agree with @mrdavidlaing, either use the array format he gave an example of, or continue using a hash but as a single key and with multiple pairs...

```
options:
  add_field:        # becomes...    add_field => {"name1"=>"value1","name2"=>"value2"}
    name1: "value1"
    name2: "value2"
```

I don't think you'll have to change anything in ERB for this `add_field` support, just how it's configured in your YAML.

axelaris:
Guys, issue with quotes was fixed by changing `<%= v %>` to `<%= v.inspect %>`. Now it's no need to use quotes inside quotes.
As for @mrdavidlaing suggestion to use array instead of hash for `add_field` - it also does work!
So now everything is looking fine.

jobs/parser/spec
```diff
@@ -33,6 +33,8 @@ properties:
   logstash_parser.filters:
     description: "The configuration to embed into the logstash filters section"
     default: ''
+  logstash_parser.inputs:
+    description: "The configuration to embed into the logstash inputs section"
```
mrdavidlaing: @axelaris, @dpb587, Last thing to do before we merge this - write some documentation.

Any idea how we can format the documentation here is the spec file so it looks good at http://bosh.io/jobs/parser?source=github.com/logsearch/logsearch-boshrelease&version=22#p=logstash_parser.outputs ?

axelaris:
@mrdavidlaing, I have an idea.
We can go forward and merge `redis` property into `logstash_parser.inputs`. So, since this property will be filled from the box with default values (Redis), it will be easier to understand how to use it.
Agree?

mrdavidlaing:
@axelaris I'm not sure I quite understand how this will work.  Can you give more details?

I don't want existing users to have to change their manifests if they are going to continue to use redis

axelaris:
Yes, the idea was to replace properties

```
redis:
   host: redishost
   key: logstash
   port: 6379
```

with

```
logstash_parser:
   inputs:
      - plugin: redis
      options:
         host: redishost
         key: logstash
         port: 6379
```

But you're right, users will have to change their manifests in this case. Apologies, I got excited when suggested this.

mrdavidlaing:
@axelaris - A cool idea; but unfortunately one which would impact all existing users even if they weren't using an alternate queue; so thus not really an option.   :(

dpb587:
If you're interested in merging behaviors, instead of dropping `redis.*` properties, we could clean the file up _and_ make this backwards compatible by:
1. Defaulting `logstash_parser.inputs` to `[ { plugin: "redis", options : {} } ]`.
2. Updating `input_and_output.conf` just inside the loop to default redis inputs with our historical, primary properties using something like...
   
   ```
   if 'redis' == input['plugin']
     input['options'] = {
       "host" => p('redis.host'),
       "port" => p('redis.port'),
       "type" => "redis-input",
       "data_type" => "list",
       "key" => p('redis.key'),
       "threads" => 8,
     } + input['options']
   end
   ```
3. Updating `input_and_output.conf` to drop the hard-coded, existing `redis { ... }` section; drop the added `an_input` and `raise`.

mrdavidlaing:
@dpb587 Now that sounds like a plan.  @axelaris - you up for implementing this?

axelaris:
Sure, I'm going to try this out!
@dpb587, thanks, as usual nice idea!

mrdavidlaing:
@axelaris - As to how to document this - another good idea from @dpb587  - could write the documentation in the `spec` ala: https://github.com/cloudfoundry-incubator/uaa-release/blob/develop/jobs/uaa/spec#L153-L165, ie:

```
logstash_parser.inputs:
    description: |
      A nested or flat hash of messages that the login server uses to display UI message
      This will be flattened into a java.util.Properties file. The example below will lead
      to four properties, where the key is the concatenated value delimited by dot, for example scope.tokens.read=message
    example:
      messages:
        scope:
          tokens:
            read: View details of your approvals you have granted to this and other applications
            write: Cancel the approvals like this one that you have granted to this and other applications
      scope.tokens.read: View details of your approvals you have granted to this and other applications
      scope.tokens.write: Cancel the approvals like this one that you have granted to this and other applications
```

axelaris:
I'm done. Sorry for delay.
Checked and confirmed to work fine with `logstash_parser.inputs` value defined and undefined.

cloudcredo-slack:
This has evolved into a really elegant solution.  It's been a pleasure
working with you on getting here.

I'll try get this merged with @dpb587 tomorrow.

Does the matching PR for logsearch-for-cloudfoundry need any changes?

axelaris:
I'm done with corresponding changes in logsearch-for-cloudfoundry.
Please check https://github.com/logsearch/logsearch-for-cloudfoundry/pull/65

mrdavidlaing:
@axelaris; great, thanks.  We'll try get them all merged and released together

dpb587:
Merged - thanks again for your work, brainstorming, and time on this!

axelaris:
Cool!
What about https://github.com/logsearch/logsearch-for-cloudfoundry/pull/65 ?

mrdavidlaing:
@axelaris - logsearch/logsearch-for-cloudfoundry#65 merged via logsearch/logsearch-for-cloudfoundry#70

Issue #72, opened 2014-08-14
Elasticsearch disk space


teddyking:
Hi there,

We (CloudCredo) have recently run into a couple of issues with logsearch regarding elasticsearch and disk space.
I can see from [Issue #36](https://github.com/logsearch/logsearch-boshrelease/issues/36) that you have had some similar problems in the past.

Is there a recommended approach to dealing with this? I've been thinking about extending our logsearch deployment(s) with [curator](https://github.com/elasticsearch/curator) but I wanted to check if you had any plans to do something similar in the future?

Thanks,
Ed

dpb587:
Hi Ed - thanks for taking the time to post! The disk space has been a common issue, and so far we've been using curator, too. Currently, our approach has been non-integrated - scheduling the curator task separately through a stable machine's cron or Jenkins.

But, we'd eventually like to see that cleanup/curator process integrated with the deployment. These are some of the implementation ideas we've internally touched on in the past, and I'd be very interested if you had other ideas or even wanted to give them a go yourself:
- curator as a BOSH errand (integrated, but doesn't inherently solve the scheduling aspect)
- a Jenkins build template which could be deployed somewhere in the network (non-integrated)
- curator as a job template with a long-running wrapper and lightweight lock (needs python dependency, probably installed on master nodes, the process would be sleeping 99% of the time between runs, use `_ttl` and elasticsearch docs for only running once if installed on multiple jobs)

So far, this issue hasn't been a priority amongst our other tasks. If you're able to make some progress on this that you can share, we'd love to see a PR. Or if you just want to chat implementation ideas, this issue will be a great place to do that, too.

dpb587:
Another monit-based idea which only curates when space is limited...

```
check filesystem xvdfsomething with path /var/vcap/store
    if space usage > 90% then exec '/path/to/curator -d <%= p('elasticsearch.curator.retain_days') %> --other-options...
```

Wouldn't handle daily calls for index optimization though.

mrdavidlaing:
I like the above idea which links curating to the actual problem trigger - lack of disk space

Monit also seems to have a CRON like scheduler, so we could also do a daily optimise like this:

```
check program elasticsearch_optimiser with
          path /var/vcap/jobs/elasticsearch/bin/optimise 
              timeout 3600 seconds
   every "* 2 * * *"
```

Since monit allows multiple check statements in the monit config for a specific job, we could have a collection of these.

teddyking:
[...]

PR #164, opened 2015-08-11
Errand that purge logs with curator

bonzofenix:


bonzofenix:
This should bring some progress on #72

Issue #138, opened 2015-06-20
Remove empty syslog fields


dpb587:
I think our standard syslog filters should drop `syslog_msgid` and `syslog_procid` when their value is `-`. It's just two extra fields that don't need to be stored or reviewed with event messages.

PR #175, opened 2015-08-26
Ignore syslog_procid + syslog_msgid when they are unused

dpb587:
If the values are `-`, they'll get ignored in the `grok`ing. If the message came from `nxlog`, we drop them along with the other fields which are known to be useless with nxlog as the forwarding transport.

Issue #138


dpb587:
@mrdavidlaing, any objections?

mrdavidlaing:
@dpb587 - Nope - I think this is a good idea.  Merge away

PR #176, opened 2015-08-31
Curator job: fix new curator job using non existing resource pool

fbonelle:
In latest release (v23.0.0), a new job called curator is using a non existing resource pool. Template generation with spiff is broken.

I have fixed spiff generation by updating the templates.

A new resource pool have been added: logsearch_curator.


dpb587:
This PR looks sane to me - https://gist.github.com/dpb587/051ad80070b10980c1b8

PR #179, opened 2015-09-11
Added Rabbit to ingestor_syslog

axelaris:
Hi guys, I've just added a Rabbit (and so on) support for **ingestor_syslog**, just like in #160


mrdavidlaing:
@axelaris - Thanks for this PR; I think its a great extension.

However, I'm concerned that we should also be adding the same features to the other ingestors (`ingestor_relp` etc).

Do you think it would be possible to use a similar `ingestor.inputs` technique that would enable us to have a single generic `ingestor` job that can replace all our existing `ingestor_*` jobs?

axelaris:
I guess it will be possible to implement the same feature into each other `ingestor_*` job.

As for merging all `ingestor_*` jobs into a single generic `ingestor` job - it will require some time to merge all templates and properties, but I bet it's also possible. But we should make sure that it still be possible to run few ingestor jobs with different input methods on a single VM (now it's possible just because job names are different). 

Anyway, reasonability of this move should be discussed with other team members.

axelaris:
Added **ingestor_relp** job. Not sure if I should add **ingestor_lumberjack** here, because I'm working with CloudFoundry and Lumberjack is not seems to be a supported protocol: https://docs.cloudfoundry.org/running/managing-cf/logging.html
Do you want me to add it without testing?

mrdavidlaing:
@axelaris, I agree;  ingestor_lumberjack shouldn't be changed.  I actually think we should remove that ingestor, since it's unreliable and no-one is using it.

mrdavidlaing:
https://github.com/logsearch/logsearch-boshrelease/issues/182 - proposal to remove ingestor_lumberjack

axelaris:
All ingestors done

jobs/ingestor_archiver/spec
```diff
@@ -27,6 +27,18 @@ properties:
...
+  logstash_ingestor.outputs:
+    description: |
+      A list of output plugins, with a hash of options for each of them. Please refer to example below.
+    example:
+      outputs:
+        - plugin: rabbitmq
+          options:
+            host: 192.168.1.1
+            user: logsearch
+            password: c1oudbunny
```
dpb587, line 39: :)

jobs/ingestor_relp/spec
```diff
@@ -9,7 +9,7 @@ templates:
   data/properties.sh.erb: data/properties.sh
   helpers/ctl_setup.sh: helpers/ctl_setup.sh
   helpers/ctl_utils.sh: helpers/ctl_utils.sh
-  config/relp_to_redis.conf.erb: config/relp_to_redis.conf
+  config/logstash.conf.erb: config/logstash.conf
```
dpb587, line 12: I appreciate the more generic file name.

jobs/ingestor_archiver/spec
```diff
@@ -27,6 +27,18 @@ properties:
   logstash_ingestor.debug:
     description: Debug level logging
     default: false
+  logstash_ingestor.outputs:
+    description: |
+      A list of output plugins, with a hash of options for each of them. Please refer to example below.
+    example:
+      outputs:
```
dpb587, line 34: Traditionally-without-knowing-the-actual-tradition, the value of `example` should be whatever the value of the property is - same format as `defaults` - so I don't think the mid-level `outputs` key should be included.

dpb587, line 34: Which I guess I missed [earlier](https://github.com/logsearch/logsearch-boshrelease/blob/develop/jobs/parser/spec#L40) - my bad.

axelaris:
Hi guys,
it's seems we are stuck here for a month.
Are you waiting for more input from my side?

mrdavidlaing:
@axelaris - I'm sorry; we've been snowed under with CF Summit Berlin.  We don't need anything more from you; will hopefully get to merging this in the next few weeks.

axelaris:
I hope you are well, guys ;-)

PR #186, opened 2015-09-24
Cleanup syslog fields

mrdavidlaing:
This PR:
1.  Ensures that `@shipper.host` is always populated with the address of the machine sending the data.
2.  When `logstash.metadata_level` == `DEBUG`, adds
   - `@ingestor.timestamp` 
   - `@ingestor.service` 
   - `@ingestor.job` 
3.  Removes `host`, `received_at`, `received_from` (since these are captured by `@shipper.host` and @ingestor.timestamp`
4.  Ensures `@shipper`, `@source` and `@ingestor` are objects (rather than strings with periods in them; which is not allowed in ES 2.0)
5.  Cleans up the `fail/syslog_standard/*` error tags
6.  Retain the `syslog_hostname` and `syslog_timestamp` fields

### DIFF
--- a/jobs/ingestor_syslog/templates/config/syslog_to_redis.conf.erb
+++ b/jobs/ingestor_syslog/templates/config/syslog_to_redis.conf.erb
@@ -25,9 +25,9 @@ input {
 }
 
 filter {
-	#Store host as @ingestor.remote_host
+	#Store host as @shipper.host
 	mutate {
-		add_field => { "[@ingestor][remote_host]" => "%{host}" }
+		add_field => { "[@shipper][host]" => "%{host}" }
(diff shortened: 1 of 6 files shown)


mrdavidlaing:
Related - 3bf496a57dd5c68d7cd453703ade6793cd62a157

When the syslog message gives us additional information about the original message host (via `[syslog_sd_params][host]`), treat the `syslog_hostname` as the `@shipper.host` and the @source.host as  `[syslog_sd_params][host]`

Specifically:
- `[@shipper][host]` => `%{[syslog_hostname]}`
- `[@source][host]` => `%{[syslog_sd_params][host]}`

PR #188, opened 2015-10-09
Fix curator resource pool

rkoster:
Resource pool: `logsearch` [does not exist anymore](https://github.com/logsearch/logsearch-boshrelease/blob/develop/templates/logsearch-deployment.yml#L38-L67).


cromega:
#176 is a better fix.

Issue #189, opened 2015-10-12
Broken `elasticsearch.indices.ttl_default` property


ghost:
The `elasticsearch.indices.ttl_default` property does not seem to work anymore.

It is still located in the elasticsearch `spec` file: https://github.com/logsearch/logsearch-boshrelease/blob/develop/jobs/elasticsearch/spec#L51
But it is not used in any of the templates: https://github.com/logsearch/logsearch-boshrelease/search?utf8=%E2%9C%93&q=ttl_default

It used to be part of the default ElasticSearch template: https://github.com/logsearch/logsearch-boshrelease/blob/v17/jobs/elasticsearch/templates/config/templates/logstash_default.json.erb#L30
But I guess it got lost along the way.

Regards,
Momchil Atanassov

dpb587:
The property was removed in #112 when we stripped deployment-specific properties, but I apparently forgot to remove it in the spec file. You can continue to configure the [TTL settings](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-ttl-field.html#_note_on_documents_expiration) with the `elasticsearch.config_options` property (and including the `indices.ttl.interval`, if needed), but also be sure to configure [`_ttl`](https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-ttl-field.html) for the specific mappings where you need it in your indices.

PR #190, opened 2015-10-13
Drop unused elasticsearch.indices.ttl_default job property

dpb587:
Issue #189


cromega:
Thanks! This property is not used any more indeed.

PR #192, opened 2015-11-05
Added logsearch_curator resource pool

axelaris:
Resource pool for `curator` job was defined as `logsearch`, but there was no such resource pool.


dpb587:
@mrdavidlaing, so there are now 3 PRs about this (#176, #188, and this). I think #176 is the most comprehensive of the 3, when you get some time to catch up on these.

cromega:
I merged #176 which renders this PR obsolete.

PR #194, opened 2015-11-15
Decrease persistent disk sizes to 40G

voelzmo:
400G seems a bit overambitious.

### DIFF
--- a/generate_deployment_manifest
+++ b/generate_deployment_manifest
@@ -7,6 +7,7 @@ infrastructure=$1
 shift
 
 if [ "$infrastructure" != "aws" ] && \
+    [ "$infrastructure" != "openstack" ] && \
     [ "$infrastructure" != "warden" ] && \
     [ "$infrastructure" != "vsphere" ] ; then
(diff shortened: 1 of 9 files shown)


poblin-orange:
True 400Go is too much.
Ideally, this value should be settable in stub manifest, so we can customize the value for each deployment

mrdavidlaing:
We'll take a PR to make this configurable; but 40GB is _waay_ too small for any production like installation.

The default needs to err on the side of too large; since ES dies horribly when it runs out of disk space.

voelzmo:
I agree that there is always the problem with productive use vs. 'just trying it out' or 'dev-usage' for templates.

Not sure I agree that the default should err on the side of being too large. People deploying this to production are probably more willing to adjust some values and go more thorough through the settings than someone who wants to 'just deploy' the ELK stack.

Even if this was configurable, which I again agree is preferable to just setting it to some value, we'd probably argue about the 'default value': Should it be 'production ready' or 'dev-quality'?

cromega:
We are working on a major overhaul of the whole template generation bit. This PR is going to be obsolete very soon.

Issue #181, opened 2015-09-14
Prevent elasticsearch from using swap


mrdavidlaing:
Although we set [`bootstrap.mlockall: true`](https://github.com/logsearch/logsearch-boshrelease/blob/develop/jobs/elasticsearch/templates/config/config.yml.erb#L1) in the elasticsearch config, it doesn't appear to be working, since I get the following WARN in the elasticsearch node logs:

```
[2015-09-14 06:33:51,181][WARN ][bootstrap                ] Unable to lock JVM memory (ENOMEM). This can result in part of the JVM being swapped out. Increase RLIMIT_MEMLOCK (ulimit).
```

and `curl http://127.0.0.1:9200/_nodes/process?pretty | grep mlockall` gives:

```
        "mlockall" : false
        "mlockall" : false
        "mlockall" : false
        "mlockall" : false
```

https://www.elastic.co/guide/en/elasticsearch/reference/current/setup-configuration.html#setup-configuration-memory seems to contain some guidance as to what is going wrong

dpb587:
I think the best first step would be adding `ulimit -l unlimited` to the control scripts, similar to how we raise the [file limit](https://github.com/logsearch/logsearch-boshrelease/blob/33d3d4e6c853504b0d87d0044d3c92f9d0166b9f/jobs/elasticsearch/templates/bin/elasticsearch_ctl#L61).

It's not a good idea to disable swap unless you have a really, really good understanding of what's happening on the VM and I don't think that's something we should try to generalize in our release. That's more something to be handled by a colocated, generic utility release responsible for configuring sysctl, /proc/sys, and other low level settings.

PR #195, opened 2015-12-14
Add ulimit for mem

axelaris:
Hi guys,
this quick fix allows to get rid of annoying warning `Unable to lock JVM memory (ENOMEM). This can result in part of the JVM being swapped out. Increase RLIMIT_MEMLOCK (ulimit).` when ElasticSearch starting


dpb587:
Looks good to me - related to and should be the fix for #181.

PR #198, opened 2016-01-25
Add openstack

axelaris:
Hi guys,
do you want to merge Openstack stuff?


templates/logsearch-infrastructure-openstack.yml
```diff
@@ -0,0 +1,63 @@
+compilation:
+  cloud_properties:
+    instance_type: (( grab meta.instance_type.compilation ))
+
+resource_pools:
+- name: elasticsearch_master
+  cloud_properties:
+    availability_zone: (( grab meta.availability_zone ))
+    instance_type: (( grab meta.instance_type.elasticsearch ))
```
cromega: Would you provide sensible defaults for these fields? Spruce does things differently from spiff, check the aws templates for guidance. Try not to use the "meta convention" where it can be avoided.

https://github.com/geofffranks/spruce

Cheers!

axelaris:
Hi @cromega, thanks for a comment.
I've checked aws templates and found the thing which I don't like. Resource pools (and instance_types) are defined both in `logsearch-infrastructure-aws.yml` and `stub.aws.example.yml`. It's looks weird.

I believe it's much better to define resource pools once, and take all variables from `meta`, which can be imported from a stub file.

As for default values, I've tried to specify them, as described in the documentation:

```
compilation:
  cloud_properties:
    instance_type: (( grab meta.instance_type.compilation || "m1.small" ))
```

However, it does not work. Not sure why:

```
$ ./scripts/generate_deployment_manifest openstack stub.yml 
1 error(s) detected:
 - $.compilation.cloud_properties.instance_type: Unable to resolve `||`: `||` could not be found in the YAML datastructure
```

mrdavidlaing:
@axelaris - Could you email me at dlaing@pivotal.io - I'd like to figure out how you can meet the team so we can work more closely on this kind of stuff.

axelaris:
Hi @cromega, @mrdavidlaing.
I've updated Spruce version, it was help to resolve the issue with '||'.
Also I've added default values to the stub.

PR #199, opened 2016-01-29
Update spruce

axelaris:
Included Spruce binaries are outdated and fails to handle `||` statement: https://github.com/logsearch/logsearch-boshrelease/pull/198
This is update to v1.0.1


cromega:
thanks!

PR #202, opened 2016-02-17
Json key name fix

hannayurkevich:
It looks like we should use another field to set `parsed_json_key` in `if_it_looks_like_json.conf`:

`if`**`[@source][service]`**`{`
`mutate { add_field =>`**`{ "parsed_json_key" => "%{[@source][service]}" }`**`}`
`} else if [syslog_program] {`
`mutate {`
`add_field => { "parsed_json_key" => "%{syslog_program}"}`
`}`
 `} else {`
 `mutate {`
`add_field =>`**`{ "parsed_json_key" => "unknown_source" }`**
`}`
 `}`

Currently field `parsed_json_key` is set from `[@source][service]`, but actually `[@source][service]` is not mentioned in any logstash config in the project. So it looks like this field is never set for any logs. As the result `parsed_json_key` is set to `unknown_source` for most cases.

From the other hand, we set field `[@source][component]` when parsing, so it can be a good candidate to use it for `parsed_json_key` population.


cromega:
@hannayurkevich: we rely heavily on @source.service, The rules that create @source.service are not in this bosh release, though. I agree it's quite confusing the way it is. I'm happy to merge in a fix to fall back to using @source.component if @source.service is not present. Could you change your code accordingly?

We will think about how to clear it up.

hannayurkevich:
Hi @cromega,

It's a good point. I've made the changes.
Now `[@source][component]` field is used only if `[@source][service]` is not set. See files changed.

Thanks!

cromega:
Great stuff, thanks a lot!

PR #203, opened 2016-02-19
Added support for configuring elasticsearch http.host property

stoyanr:
We would like to be able to configure the elasticsearch `http.host` property. This allows us to better control the external visibility of the HTTP service exposed by elasticsearch. For example, by specifying "127.0.0.1" we can make it visible only by other processes running on the same host, e.g. a reverse proxy.


cromega:
good one, thanks!

PR #205, opened 2016-02-19
Added support for configuring redis memory settings

stoyanr:
We would like to be able to configure the Redis memory settings for the `queue` job, in particular the 2 properties `maxmemory` and `maxmemory-policy`. We had actual issues with Redis consuming too much memory in some of our deployments. Without being able to specify these settings, we wouldn't be able to solve them.

### DIFF
--- a/jobs/queue/spec
+++ b/jobs/queue/spec
@@ -18,8 +18,9 @@ properties:
     description: Name of queue to pull messages from
     default: logstash
   redis.maxmemory:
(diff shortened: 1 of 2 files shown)


cromega:
Would you make it clear in the job spec that the NINETY_PERCENT_OF_RAM is a kind of magic value? Just in case someone wonders how it makes a useable configuration value in a deployment.

stoyanr:
I added that now, plus a bit more information and an example.

cromega:
thanks!

PR #207, opened 2016-02-22
Remove duplicate update definitions in jobs

mtekel:
YAML permits only one definition of hash. The 2nd definition is over-
riding the 1st one, which then never takes effect. Remove the 1st
definition to remove potential confusion.


mrdavidlaing:
@mtekel - thank you for this submission - we'd like to merge it; but we don't seem to have a signed Contributors License Agreement on record for you.

Please could you print & sign the relevant CLA from https://github.com/logsearch/licensing#signing-the-contributors-license-agreement and forward to david.laing@labs.cityindex.com

Thanks!

D

PR #208, opened 2016-02-23
Don't run ELS on parser

mtekel:
There is already elasticsearch deployed as a part of parser job. No
need to run another elasticsearch on the same VM. Moreover, ELS is
configured to consume 46% of the VM memory, which in case of 2 ELSes
means that the VM runs out of memory, starts swapping heavily and
becomes unresponsive.

### DIFF
--- a/templates/logsearch-jobs.yml
+++ b/templates/logsearch-jobs.yml
@@ -175,7 +175,6 @@ jobs:
   release: logsearch
   templates:
   - {name: parser, release: logsearch}
-  - {name: elasticsearch, release: logsearch}
   resource_pool: parser
   instances: 1
   networks:


cromega:
@mtekel: the change you are trying to undo came from this commit: https://github.com/logsearch/logsearch-boshrelease/commit/8d50e5e5933d84841eb033c76da60dc4041ffb17

We got rid of the nested ElasticSearch and changed the release so we use the same ElasticSearch job as used in the other jobs.

Could you explain what you have done that resulted in 2 ElasticSearch processes running on your parsers?

mtekel:
Deployed v200. I haven't checked if it was fixed since (which is was). When are you planning to make a new release?

cromega:
@mtekel: deployed v200 or upgraded to v200? If it's a fresh deployment and you actually have 2 ES processes running on the parsers then it's a massive bug in the release and we will fix it ASAP once we figured out how it happened. If that's the case, would you send me your deployment manifest?

cromega:
@mtekel: ah, I see it is fixed now, good to hear. We will probably cut a new release very soon.

Issue #209, opened 2016-02-26
Drain scripts waiting for green cluster status are dangerous


voelzmo:
I recently discovered that you added drain scripts for elasticsearch at some point in time. How they're implemented is kind of dangerous, and here is why:
- your drain scripts wait for a `green` elasticsearch cluster status
- the agent executes drain scripts _forever_, there is _no_ timeout

So guess what happens when you want to repair a broken elasticsearch cluster with a `bosh deploy`? The drain hangs forever, nothing moves.

I know (now) that I can disable the drain scripts from the manifest. That didn't really do it for me because I only realized it when I had it already deployed. I know I can do `bosh stop --skip-drain` and then re-deploy, which I eventually did.

I'm not sure which cases you're covering with the scripts, maybe a scale-down without dataloss? Seems like the current idea of the drain has some unwanted side-effects you want to think about or document thoroughly.

ghost:
To add to @voelzmo , it seems also that the drain script disables allocations, which are reenabled only once a node is back up (in the ctl script). This means that if you downscale the cluster (where the node will not return) and are not aware of this, the cluster will remain locked and unusable.

I believe the idea behind this script was to prevent a massive rellocation of data during an update (e.g. stemcell update) of the cluster.

cromega:
@voelzmo, @momchil-sap: we are ware of the problems.

We completely reworked the data node upgrade process, the changes are waiting for a new feature to be added to BOSH which is expected to be released with one of the next few releases, we are hoping within a few weeks.

voelzmo:
@cromega Thanks for explaining the rationale behind this. Out of curiosity: which feature would that be that you're waiting for?

cromega:
@voelzmo: the ability to run post-deploy scripts

voelzmo:
Ah nice, that is on the `develop` branch already, so the next release should probably have it :)

mrdavidlaing:
@voelzmo - The next release - v201 - will

(a) Disable shard allocations at the beginning of a deploy
(b) Have an errand - `bosh run errand enable-shard-allocation` to re-enable then after a deploy has finished

As soon as BOSH releases with post-deploy scripts, (b) will be turned into a  post-deploy scripts.

Hope that helps; and thanks for your thoughts!

dpb587:
Does fully disabling allocations prevent indices from being created if log
messages with new dates/times come through, potentially causing data loss
until re-enabling?

[...]

PR #216, opened 2016-03-14
Drain cannot be disabled via the manifest

adamstegman:
It looks like a side effect of the fix for #209 was that it's no longer possible to set `elasticsearch.drain` to `false`. Was this intentional? If not, here's a quick PR to replace it.


cromega:
@adamstegman: yes, it is intentional. Relying on a deployment property to turn the drain feature on and off is and has always been a broken mechanism.

Deploying with `elasticsearch.drain: false` does not actually turn off this feature, it just updates the drain script with an early `exit 0`.

Imagine you have the drain turned on but you actually want to deploy without it because something breaks because of it. First you would need to get a deploy through without problems in order to update the drain scripts with the new content.

Drain is supposed to be controlled via the appropriate BOSH cli flag:

`bosh deploy --skip-drain elasticsearch_data`

PR #224, opened 2016-04-14
Add support for elasticsearch routing

bradylove:


mrdavidlaing:
@bradylove, @apoydence - could you provide a little context for how this feature will be used - will be helpful for writing the release notes.

poy:
We want to shard on the uuid of each message.

bradylove:
@mrdavidlaing to add to what @apoydence said, we are routing on the application id of each message. We do this to help keep searches performant. When we are performing a search we are always filtering by the application id so this will limit the number of shards that need to perform the query.

mrdavidlaing:
@bradylove, @apoydence, The relevant section in the Elastic documentation is here - https://www.elastic.co/guide/en/elasticsearch/guide/current/user-based.html - it deals with forums rather than apps; but its conceptually the same.

Whilst I can see that this will improve search response times; my gut instinct is that our bottleneck is going to be ingesting rather than searching; and I'm concerned that sending all data from a single chatty app to a to a single shard might be less performant that spreading the indexing load over many shards.

We'll only know which set of trade-offs is best after we've done some proper load testing on PWS; so I'll motivate to get this merged.

@cromega - How will enabling index routing affect our existing setup?  Specifically:
1.  What happens for logs that don't have the routing key ( or can we generate one for them )
2.  Does Kibana support querying data that has been "routed".  (Can't think why not; but worth checking)

mrdavidlaing:
An interesting read on optimising indexing performance - https://www.elastic.co/guide/en/elasticsearch/guide/current/indexing-performance.html.

Unfortunately it doesn't discuss the impact of custom routing to indexing performance, but it does have some interesting other knobs to twiddle if we have indexing perf issues.

kirederik:
@apoydence, @mrdavidlaing, @bradylove 

1. Indexing without routing may lead to undesired results such as objects getting stored in multiple shards, according to the documentation. Workaround would consist of setting the routing field to required in the index mapping, we would also have to modify our parsing rules to make up some routing value if it is missing. In case of a bug in the parsing rules we would reject lots of logs.

2. Kibana does not support searching with custom routing, the workaround would be to create an index alias with the routing configured, which we can't do at the moment without further changes to the release.

cromega:
@apoydence, @mrdavidlaing, @bradylove 
1. Indexing without routing may lead to undesired results such as objects getting stored in multiple shards, according to the documentation. Workaround would consist of setting the routing field to required in the index mapping, we would also have to modify our parsing rules to make up some routing value if it is missing. In case of a bug in the parsing rules we would reject lots of logs.
2. Kibana does not support searching with custom routing - https://github.com/elastic/kibana/issues/3298 - , the workaround would be to create an index alias with the routing configured, which we can't do at the moment without further changes to the release.

mrdavidlaing:
@cromega,  I'd like to do some experiments with the latest version of Kibana to verify how it breaks so we can document accordingly.

mrdavidlaing:
Ok, experiments done.

tldr; Kibana works fine searching against indexes that have had data routed to specific shards.  However, it doesn't add the routing key to its queries; so there is no search performance improvement from a Kibana perspective.   No reason not to accept the PR; although it doesn't seem to make sense to use it when Kibana is your UI.

One unrelated thing I noticed in my experiment was that depending in the routing key you choose; you can easily end up with highly unbalanced shard load.  In my test, I configured logstash with:

```
elasticsearch {
    routing =>  "%{[@source][vm]}"
    index => "logstash-with-routing-%{+YYYY.MM.dd}"
   ...
}
```

Note how the `logstash-with-routing-2016.04.22 3` shard ended up with nearly 10 times as much data as the `logstash-with-routing-2016.04.22 0` shard

```
logstash-with-routing-2016.04.22 3 r STARTED   1588   1.1mb 10.0.16.62 elasticsearch_data-partition-2724347eaea24de34113/3
logstash-with-routing-2016.04.22 3 p STARTED   1588   1.3mb 10.0.16.61 elasticsearch_data-partition-2724347eaea24de34113/2
logstash-with-routing-2016.04.22 2 p STARTED    109 211.6kb 10.0.16.60 elasticsearch_data-partition-2724347eaea24de34113/1
logstash-with-routing-2016.04.22 2 r STARTED    109 211.6kb 10.0.16.61 elasticsearch_data-partition-2724347eaea24de34113/2
logstash-with-routing-2016.04.22 4 p STARTED    135 195.5kb 10.0.16.59 elasticsearch_data-partition-2724347eaea24de34113/0
logstash-with-routing-2016.04.22 4 r STARTED    135 195.5kb 10.0.16.62 elasticsearch_data-partition-2724347eaea24de34113/3
logstash-with-routing-2016.04.22 1 p STARTED   1161 701.7kb 10.0.16.62 elasticsearch_data-partition-2724347eaea24de34113/3
logstash-with-routing-2016.04.22 1 r STARTED   1161 756.9kb 10.0.16.60 elasticsearch_data-partition-2724347eaea24de34113/1
logstash-with-routing-2016.04.22 0 p STARTED     21    66kb 10.0.16.59 elasticsearch_data-partition-2724347eaea24de34113/0
logstash-with-routing-2016.04.22 0 r STARTED     21    66kb 10.0.16.60 elasticsearch_data-partition-2724347eaea24de34113/1
```

For comparison with no routing key specified ( the default ), data is spread pretty evenly across shards

So, choosing a routing key is pretty important - all logs must have it; and its values should be evenly distributed.

mrdavidlaing:
@bradylove, @apoydence, Sorry for the delay; this change has now been merged.

PR #228, opened 2016-05-12
elasticsearch_data job should use dynamic IPs

syslxg:
when I scale out the es cluster, I don't want to worry about finding extra static IPs, just bump the instance number and deploy.


mrdavidlaing:
Good to merge; @syslxg is covered by the Pivotal CLA

kirederik:
Thanks!

PR #231, opened 2016-05-26
Limit the number of messages that can be sent to the queue

mrdavidlaing:
Newer version of the [`logstash redis output come with congestion settings`](https://www.elastic.co/guide/en/logstash/current/plugins-outputs-redis.html#plugins-outputs-redis-congestion_threshold) that can be used to prevent logstash from overloading redis.

This PR adds configuration options to the ingestor_syslog job that limit the number of messages it will place on the queue to 100,000

### DIFF
--- a/jobs/ingestor_syslog/spec
+++ b/jobs/ingestor_syslog/spec
@@ -55,6 +55,12 @@ properties:
   redis.key: 
     description: Name of queue to pull messages from
     default: logstash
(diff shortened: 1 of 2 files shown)


.bundle/config
```diff
@@ -0,0 +1 @@
+--- {}
```
dpb587: 🐲

PR #232, opened 2016-05-31
Fix hash in relp input add_field

tushar-dadlani:


mrdavidlaing:
This commit is good to merge, since @tushar-dadlani is covered by the Pivotal CLA

PR #235, opened 2016-06-13
Ensure elasticsearch is receiving requests in ctl script before exiting

bradylove:
Sometimes when doing a bosh deploy we have ran into an issue where the master node will deploy and report back healthy, then bosh will continue on to deploy the data nodes which will fail because the master is not accepting connections yet.

This change ensures that the elasticsearch node is accepting connections before continuing


cromega:
@bradylove: interesting. Are you deploying with some funky settings perhaps? Which IAAS?

I have never seen this happening but I can definitely see the benefit of having such a check.

bradylove:
@cromega We ran into this issue when deploying to AWS, It doesnt happen often but is annoying when it does. 

Unfortunately we didn't save any logs from the data nodes when this happened.

bradylove:
@cromega We are currently unable to reproduce the issue that was meant to fix. I am going to close the PR for now. We are going to explore having multiple master nodes which from the sound so it would resolve this problem and will solve some other issues for us.

PR #237, opened 2016-07-04
Fix generate_deployment_manifest usage

Infra-Red:
Hi!
I updated `generate_deployment_manifest` to write in stdout in case of misspelling or wrong usage, it was annoying to find `echo "usage: ./generate_deployment_manifest <aws|openstack|warden|vsphere> [stubs...]"` inside of my deployment manifest.
Also i can propose to delete Spruce binary from logsearch-boshrelease repo and instead of it add simple check into `generate_deployment_manifest` script for Spruce installation. 
@mrdavidlaing What is your opinion on this? If you agree, than i can add additional commit to current PR.
Thanks!

### DIFF
--- a/scripts/generate_deployment_manifest
+++ b/scripts/generate_deployment_manifest
@@ -12,7 +12,7 @@ if [ "$infrastructure" != "aws" ] && \
     [ "$infrastructure" != "openstack" ] && \
     [ "$infrastructure" != "warden" ] && \
     [ "$infrastructure" != "vsphere" ] ; then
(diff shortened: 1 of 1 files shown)


cromega:
@Infra-Red 

The change to the script is 👍 , thanks!

why do you want to remove spruce? I like the idea that I can run the scripts without having to fetch external tools

Infra-Red:
@cromega I believe that this will make repo a bit lighter. and in our BOSH world you simply cannot live without installed `spruce` and `spiff` 😉

PR #239, opened 2016-07-15
Added Elasticsearch web access

axelaris:
Hi,
this PR is adding ability to access Elasticsearch interface from the browser. This is very useful for Kopf, for example.
Here is an example of properties for `ls-router` job for using it:

```
  properties:
    haproxy:
      elastic:
        inbound_port: 8888    # Must specify the port to enable access to Elasticsearch
        backend_servers:
        - (( grab jobs.elasticsearch_master.networks.default.static_ips.[0] ))
        auth:
          username: user
          password: pass
```

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -82,6 +82,37 @@ backend cluster_monitor
     server kibana<%= index %> <%= ip %>:<%= p("haproxy.cluster_monitor.backend_port") %> check inter 1000
     <% end %>
 
+<% if p('properties.haproxy.elastic.inbound_port') %>
+frontend elastic-in
+    mode http
+
+    <% if p('properties.haproxy.ssl_pem') %>
+    bind :<%= p("haproxy.elastic.inbound_port") %> ssl crt /var/vcap/jobs/haproxy/config/cert.pem no-sslv3
+    <% else %>
+    bind :<%= p("haproxy.elastic.inbound_port") %>
+    <% end %>
+
+    <% if p('haproxy.elastic.auth.username') != "" %>
+    acl logged_in http_auth(elastic-user)
+    http-request auth realm "Elasticsearch" if !logged_in
+    <% end %>
+
+    default_backend elastic
+
+backend elastic
+    mode http
+    balance roundrobin
+    <% p("haproxy.elastic.backend_servers").each_with_index do |ip, index| %>
+    server elastic<%= index %> <%= ip %>:9200 check inter 1000
+    <% end %>
+
+<% if p('haproxy.elastic.auth.username') != "" %>
+userlist elastic-user
+    user <%= p('haproxy.elastic.auth.username') %> insecure-password <%= p('haproxy.elastic.auth.password') %>
+<% end %>
+
+<% end %>
+
 <% if p('haproxy.kibana.auth.username') != "" %>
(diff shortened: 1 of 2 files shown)


cromega:
@axelaris: It would be a very simple solution to the problem but it would potentially open up a door to the cluster with full control over the data.

If you cared to take some time and to dig into haproxy configuration and look if there is a way to configure the proxy to only let certain requests through we would consider implementing it.

cromega:
My alternative suggestion is a proxying plugin for kibana that either creates a new route or extends the `/elasticsearch` one in ways to make kopf work.

Also, keep in mind that kopf itself can be used to drop all indexes.

mrdavidlaing:
@cromega,

Given that this direct Elasticsearch access requires HTTP Auth; we do at least have some "protection".  But I'd also sleep easier if it only allowed READ access.

Back in the Kibana 3 / ES 1.7 timeframe we had a nginx config that implemented READ only access; so that might be a starting point - https://github.com/logsearch/logsearch-boshrelease/blob/v23.0.0/jobs/api/templates/config/nginx.conf.erb - although it was a bit of a game of whack-a-mole as ES got / depreciated APIs.

@axelaris, some notes on how we access ES currently.
1.  You get very limited ES access through Kibana - enough to access the `_cat` apis for example: `https://<ls-router>:<haproxy.kibana.inbound_port>/elasticsearch/_cat`
2.  We normally use a SSH tunnel to access KOPF - `ssh vcap@bosh-ip -L9200:<es-master-ip>:9200` then `http://localhost:9200/_plugin/kopf`

axelaris:
Hi guys,
thank you for suggestions!
However, the point was to get the full featured access to Kopf, not limited by `read` operations. Because sometimes we want to reindex, change settings or even delete index.

And there are 2 levels of protection implemented:
1. Access to Elasticsearch will not be enabled until administrator manually add `haproxy.elastic.inbound_port` option to the manifest. This option is not specified by default. Additionally, I can add a warning message here, stating that user should understand what he is doing.
2. Basic HTTP auth allow limit access to administrators only. And this password is different from Kibana one.

@mrdavidlaing, thank you for SSH tunnel example, but anyway it is not so comfortable as direct HTTP access. It is not dying periodically because of TCP timeouts ;-)

mrdavidlaing:
@axelaris - could you add the warning to the spec as suggested.

@cromega; with the feature defaulting to off and the operator reading a warning when switching it on; I'm for this PR to be merged.

Also this is good to merge from a CLA perspective

jobs/haproxy/templates/haproxy.config.erb
```diff
@@ -82,6 +82,37 @@ backend cluster_monitor
     server kibana<%= index %> <%= ip %>:<%= p("haproxy.cluster_monitor.backend_port") %> check inter 1000
     <% end %>
 
+<% if properties.haproxy.elastic.inbound_port %>
```
cromega: just to double check: is this line not going to raise an error if the property is not set?

axelaris: Yes, it should not raise any error here.

axelaris:
Hi @mrdavidlaing, @cromega,
I've just added a warning message to the `spec` file.
Could you please review and confirm that everything is Ok from your perspective.

cromega:
cool thanks!

PR #250, opened 2016-09-16
WIP: Implement logic to send service logs to syslog

keymon:
# Motivation

Several of the CF and Diego platform services send logs to the local syslog, tagged as `vcap.*`. . The [metron agent from loggregrator configures the local syslog to send the logs to metron](https://github.com/cloudfoundry/loggregator/blob/develop/jobs/metron_agent/spec#L29) so that later can be sent to a centralised logging service. 

In some cases it is the service application itself who sends the logs to syslog, but it seems to be a common pattern across CF and Diego services to [let the service log to stdout/stderr, and redirect that output to local files or syslog](https://github.com/cloudfoundry/diego-release/blob/develop/jobs/bbs/templates/bbs_as_vcap.erb#L91) using the [`logger` command](http://linux.die.net/man/1/logger). 

Not all the services implement this feature, and instead they only log to local files. In order to centralise all the logs, so we will add the logic required to send the logs of the missing services. 

We will implement the _log to stdout and redirect  to `logger` pattern_ whenever is possible.
## Implementation in logsearch bosh release

We want to implement the same idea in all the logsearch services.

For that we we add several properties: `<service>.log_to_syslog`. When true, we redirect the stdout and stderr in the ctl script into syslog following the standard vcap pattern, in addition to files.

The only exception is `redis` in the `queue` job. In that case, we use redis capabilities to send logs to syslog, to avoid complex setups deamonising redis.
## How to review?

Enable the `*.log_to_syslog` and deploy. Logs should be sent to syslog. 

If metron agent is installed locally, the logs will be redirected to metron agent the loggregrator.
## Open questions

This PR is open to debate for several open questions:
- Is OK to add a specific property to enable this feature for each service? Shall we use one unique property? or should this be enabled by default?
- Is the tag `vcap.*` the most appropriate one? This is not a CF official release. But the generic tag vcap.\* might be handy to refer to a service deployed with bosh.
- `tee` and `logger` run as `root`. Is that fine?


cromega:
@keymon We had a similar requirement a while ago and we came up with the syslog_forwarder job that can be added as a job template to any VM, it creates a configuration file for rsyslogd instructing it to pick up the contents of certain files and send them to a remote syslog server.

it's configured like [this](https://github.com/logsearch/logsearch-boshrelease/blob/develop/templates/logsearch-jobs.yml#L145).

Is this mechanism not sufficient for you?

keymon:
Yep, we were considering do something like that, but we wanted to get the feedback from the community first, that is why we did this PR.

In the CF/Diego ecosystem they are doing the `tee+logger` redirect approach for all the services. But we were not sure for 3rd party community releases like this one.

Your approach is good for me, and if this is the way you want to proceed, feel free to close this PR. But if you prefer follow the pattern of the CF/Diego ecosystem, we can iterate on this PR.

Thx!

cromega:
@keymon the reason `syslog_forwarder` gets my vote is because its an out-of-box functionality, it has no external dependencies and it's easy to simply not deploy it if it's not needed. This was the driving logic behind its implementation.

However, I generally like loggregator and I certainly see some merit in using a logging infrastructure that works similarly to CF/Diego so I'm happy to debate it.

I see one big difference between the 2 approaches and it is the log destination. syslog_forwarder was designed to make it easy to ship the logs from the logsearch VMs to the cluster monitor VM, which, just to make it clear, is not the same ElasticSearch cluster as the "main" logsearch cluster. Whereas if the logs are picked up by the loggregator pipeline (I assume it's the CF loggregator pipeline) they will end up where all the other deployment logs are.

Does that make sense? Please explain if I misunderstood your use-case.

keymon:
> syslog_forwarder was designed to make it easy to ship the logs from the logsearch VMs to the cluster monitor VM, which, just to make it clear, is not the same ElasticSearch cluster as the "main" logsearch cluster. Whereas if the logs are picked up by the loggregator pipeline (I assume it's the CF loggregator pipeline) they will end up where all the other deployment logs are.

That is actually a good point. Currently we use loggregator and the same logsearch for everything. I guess we will check the local files in case of trouble in the logsearch cluster itself.

For me is fine and makes sense. I will close the PR.

Thank you!

